PR #203, opened 2022-06-06
Fix memory management to ensure proper component lifecycles

mmaatttt:
While updating Democues to allow re-initializing the SDK with different accounts, I noticed that the old instances were being kept in memory when they shouldn’t. It's not a huge problem for most use cases of the SDK since more often than not it would just be a singleton-ish instance, but it’s still best to be tidy.

Digging into it, our dependency injection was inadvertently a rats nest of memory retain cycles. The simplest example is `Appcues` owns `DIContainer`, but `Appcues` is also registered in `DIContainer` and so you end up with a retain cycle: `Appcues<-->DIContainer`.

The fact that we had `Appcues: AnalyticsPublishing` also caused issues because anything that referenced `AnalyticsPublishing` would also create a retain cycle. I've fixed this by creating `AnalyticsPublisher` instead, since it didn't have to be the `Appcues` class.

Maybe the biggest change is moving the `sessionID` from `SessionMonitor` to `Appcues`. The fact that the thing *monitoring* the session was also storing the session *state* meant there were a bunch of components the depended on `SessionMonitor` just to be able to read the session ID. It made the dependency graph messy and just felt like suboptimal separation of concerns in general (`DeeplinkHandler` and `AutoPropertyDecorator` relying on `SessionMonitor` is a bit weird). More critically, moving `sessionID` to `Appcues` eliminated a cyclical dependency between `AnalyticsPublisher` and `SessionManager` where the init for each required resolving the other.

Other memory issues were fixed with a few `weak` references and there was a memory leak in the debugger UI so I fixed that too. (The experience UI is fine because I've been conscious of that as we've built out.)

Finally, I also modified the debugger to ping the health endpoint only once expanded instead of on init. This wasn't a memory leak, but it did make testing trickier. And on the topic of testing, I've added a new test that inits the SDK and ensures everything goes away as expected. It's essentially just a sanity check since it can't guarantee there are no memory issues, but it's still helpful.

## New dependency graph

depencency-graph


Sources/AppcuesKit/Presentation/Debugger/UIDebugger.swift
```diff
@@ -38,7 +38,7 @@ internal class UIDebugger: UIDebugging {
     private let config: Appcues.Config
     private let storage: DataStoring
     private let notificationCenter: NotificationCenter
-    private let analyticsPublisher: AnalyticsPublishing
+    private weak var analyticsPublisher: AnalyticsPublishing?
```
iujames, line 41: curious why this one needs to be weak? Is it because the publisher holds a reference to the debugger in the subscribers?  I _thought_ there was some way to use `weakObjects()` or something like that to have a list of subscribers held weakly (like NotificationCenter does). But I'd have to refresh myself on that one, don't totally remember details.  Just thinking about possible options to avoid mistakes down the line perhaps.

iujames (approved):
awesome cleanup! love it.  could you link to that dependency graph update or just paste the image in here too? Nice to have that reference.

Sources/AppcuesKit/Presentation/Debugger/UIDebugger.swift
mmaatttt: >Is it because the publisher holds a reference to the debugger in the subscribers?
Exactly that.

It would be an option for AnalyticsPublisher to weakly hold the decorators and subscribers. A `NSHashTable` would let you do the `weakObjects()` or make the array type a simple wrapper class that holds weak references. This seems like it would be a good next step so I can do a follow up PR with that.

PR #204, opened 2022-06-06
Misc touchups

mmaatttt:
- 🚨 Fix lint warnings
- 📝 Fix documentation error
  - Helpfully enough, the objc interface flagged the wrong param name
- 📝 Add XCFramework installation instructions
- 🍏 Clean up podspec
- 🍏 Update swift tools version
  - This shouldn't matter at all, but a bit more backwards compatibility can't hurt
- 🐛 Fix bundle referencing from XCFramework
  - The `1.0.0-beta.1` XCFramework couldn't find the debug icon
- 📝 Update README image path to fix display outside of github
  - This should fix the broken image on the [CocoaPods listing](https://cocoapods.org/pods/Appcues)

### DIFF
--- a/Appcues.podspec
+++ b/Appcues.podspec
@@ -1,29 +1,16 @@
-#
-# Be sure to run `pod lib lint Appcues.podspec' to ensure this is a
-# valid spec before submitting.
-#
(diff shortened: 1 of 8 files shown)


iujames (approved):
👏

PR #209, opened 2022-06-08
Update SPI doc target

mmaatttt:
Trying this before opening an issue for the [missing docs](https://swiftpackageindex.com/appcues/appcues-ios-sdk/main/documentation/appcueskit) on the Swift Package Index.

The xcodebuild command for docs requires the _scheme_ not a target, so perhaps that's what's wrong

```sh
$ xcodebuild docbuild -scheme Appcues -sdk iphoneos -destination generic/platform=iOS
```

### DIFF
--- a/.spi.yml
+++ b/.spi.yml
@@ -2,4 +2,4 @@ version: 1
 builder:
   configs:
     - platform: ios
-      documentation_targets: [AppcuesKit]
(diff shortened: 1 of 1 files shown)


iujames (approved):
ah, `documentation_targets` feels not the best name then...

PR #210, opened 2022-06-14
Add support for closing experience when continue action is used on last step

mmaatttt:


Sources/AppcuesKit/Presentation/ExperienceRendering/ExperienceStateMachine+State.swift
```diff
@@ -131,6 +131,13 @@ extension ExperienceStateMachine.Transition {
         _ experience: Experience, _ stepIndex: Experience.StepIndex, _ package: ExperiencePackage, _ stepRef: StepReference
     ) -> Self {
         guard let newStepIndex = stepRef.resolve(experience: experience, currentIndex: stepIndex) else {
+            if stepRef == .offset(1) && experience.stepIndices.last == stepIndex {
+                return .init(
+                    toState: .endingStep(experience, stepIndex, package),
+                    sideEffect: .continuation(.endExperience(markComplete: true))
```
iujames, line 137: technically markComplete not required right? since this would get standard handling of "ended on last step so it was complete".  Only thing that caught me eye slightly different than android, but no issue i think!

iujames (approved):
did work on retreat ✅

Sources/AppcuesKit/Presentation/ExperienceRendering/ExperienceStateMachine+State.swift
mmaatttt: The Action require the parameter and the SDK in other places sets it according to the expected behaviour

PR #213, opened 2022-06-27
Fix memory leak in sticky content trait

mmaatttt:
When checking the debugger dismiss behaviour in Democues, I noticed that after the one modal displayed, the debugger wasn't dismissing properly, indicating a memory leak. Turns out it was the sticky content in that modal: the `onSizeChange` closure property of `StickyHostingController` was capturing the reference to the `ExperienceStepViewController` which in turn owned the `StickyHostingController` as a child controller.


iujames (approved):
nice catch!

PR #214, opened 2022-06-27
Idea for publishing internal SDK events to a delegate

iujames:
I'm learning more about integrations and discovering that it seems we will have an issue here if we don't allow _something_ for customers to be able to listen to internal events and re-broadcast to things like Heap, Amplitude, Mixpanel, Google Analytics.

Went down the path of quick prototyping an idea to see how it works out, really just a few minutes, so posting that idea here for feedback and discussion.  It's functional, but really not tied to it and open to other ideas for sure.

This adds an `AppcuesAnalyticsDelegate` you can register on the main `Appcues` instance.  Just a hook to get event name and props when internal events flow through.

now there are two delegates on the class, so renamed the other one `experienceDelegate` 🤷‍♂️


mmaatttt:
Sample Amplitude implementation:

```swift
import AppcuesKit
import Amplitude

extension Amplitude: AppcuesAnalyticsDelegate {
    public func didTrack(analytic: AppcuesAnalytic, value: String?, properties: [String : Any]?, isInternal: Bool) {
        // Filter out any analytics we're not interested in passing along
        guard let value = value else { return }

        logEvent(value, withEventProperties: properties)
    }
}

// somewhere after initializing both SDKs:
appcues.analyticsDelegate = Amplitude.instance()
```

I also looked at Mixpanel, which is slightly more frustrating to integration since their `Properties` aren't `[String: Any]` but rather `[String: MixpanelType]` which is a [subset](https://mixpanel.github.io/mixpanel-swift/Protocols/MixpanelType.html) (similar to what we have, actually). That would go something like this:

```swift
import AppcuesKit
import Mixpanel

extension MixpanelInstance: AppcuesAnalyticsDelegate {
    public func didTrack(analytic: AppcuesAnalytic, value: String?, properties: [String : Any]?, isInternal: Bool) {
        // Map general properties dictionary to allowed Mixpanel values
        let mixpanelProperties: Properties? = properties?.reduce(into: [:]) { result, element in
            if let value = element.value as? MixpanelType {
                result?[element.key] = value
            }
        }
        
        track(event: value, properties: mixpanelProperties)
    }
}

// somewhere after initializing both SDKs:
appcues.analyticsDelegate = Mixpanel.mainInstance()
```

mmaatttt (approved):

PR #215, opened 2022-06-28
Add documentation compilation script

mmaatttt:
The `docs.sh` script compiles the DocC files and pushes them to the `gh-pages` branch.

I added a sanity check if it's run from a branch other than `main`. When run from the release script, the branch will always be `main`, but I wanted to allow for ad-hoc updates if we need to.


iujames (approved):
awesome!

PR #218, opened 2022-06-30
Very minor documentation touch up

iujames:
Came across this one that was worded a bit strangely, while reviewing Android side.  I _think_ the updated wording matches the intention.


mmaatttt (approved):
Yes, thank you!

PR #219, opened 2022-07-04
Ensure the completion block is called when continue from the last step

mmaatttt:
From [sc-38212: iOS SDK doesn't update property from button action](https://app.shortcut.com/appcues/story/38212/mobile-ios-sdk-doesn-t-update-property-from-button-action):

> It appears that this is an issue only on the last step of an experience. The profile details are successfully updated if the action is on the first step of a two step experience.
> The issue stems from sc-37520. The first action on the button is `@appcues/continue` and that tells the SDK "go to the next step, and once we're there, continue on to any additional actions". However, if the experience dismisses, we never see another step so the `@appcues/continue` action doesn't complete and we just wait forever to get to the `@appcues/update-profile` action.
> To summarize, the issue is actually with `@appcues/continue`, because currently any action after that on the last step of an experience won't be executed.

### DIFF
--- a/Sources/AppcuesKit/Presentation/ExperienceRendering/ExperienceRenderer.swift
+++ b/Sources/AppcuesKit/Presentation/ExperienceRendering/ExperienceRenderer.swift
@@ -111,6 +111,14 @@ internal class ExperienceRenderer: ExperienceRendering {
             case .success(.renderingStep):
                 DispatchQueue.main.async { completion?() }
                 return true
+            case .success(.idling):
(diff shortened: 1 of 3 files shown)


iujames (approved):
nice find and fix!

A lot of this just seemed a lot simpler on Android side due to the nature of coroutines and suspend functions - ceding control back to caller at "resting points" in the state machine.  I wonder if at some point it would be worth a look at async/await support here for similar, and whether that would simplify anything in the code.

PR #223, opened 2022-07-14
Add support for compiling on XCode 14

iujames:
Just one minor change needed that is technically more correct on older versions as well, from what I understand based on posts here https://stackoverflow.com/a/51250282


mmaatttt (approved):
Will miss the one line workaround, but this is better. Thanks!

PR #224, opened 2022-07-20
Add centralized action queue and support for post-experience actions

mmaatttt:
This PR does two things:
1. Update `ActionRegistry` to be have a queue for all experience actions that need to be processed. By itself this changes nothing.
2. Experience actions are now mapped and handled by appending to the `ActionRegistry` queue.

The result of this is that an experience that's dismissed via a button will handle all the button actions in the queue and then the experience actions afterwards. See the example below.

## Implementation

Moving action handling to `ActionRegistry` cleaned up a bunch of messy closure passing that I was doing before and overall it's a substantially cleaner solution.

The experience actions then just need to add to the `ActionRegistry` queue. One approach was to create an `ExperienceStateObserver` and look for the `endingExperience` state, but introducing a new `processActions` side effect into the state machine proved simpler.

## Example

This is an experience that has a button set to trigger a flow and then the experience has a redirect and a content trigger. The button flow is briefly shown before the deep link is applied and the content from the experience settings is shown.

Screen Shot 2022-07-20 at 11 46 22 AM
Screen Shot 2022-07-20 at 11 46 30 AM

https://user-images.githubusercontent.com/845681/180025949-68b12504-89dd-46e8-a780-04088da0c8c9.mp4

## Diagram Changes

The state machine has a new side effect (in orange)
updated-state-machine

The dependency graph has a new dependency (in orange)
updated-dependency-graph

### DIFF
--- a/Sources/AppcuesKit/Presentation/Actions/ActionRegistry.swift
+++ b/Sources/AppcuesKit/Presentation/Actions/ActionRegistry.swift
@@ -14,6 +14,11 @@ internal class ActionRegistry {
 
     private var actions: [String: ExperienceAction.Type] = [:]
 
+    private var isProcessing = false
+    private var actionQueue: [ExperienceAction] = [] {
+        didSet { processFirstAction() }
+    }
+
     private weak var appcues: Appcues?
 
     init(container: DIContainer) {
@@ -42,16 +47,31 @@ internal class ActionRegistry {
         actions[action.type] = action
     }
 
-    func actionClosures(for actionModels: [Experience.Action]) -> [(@escaping Completion) -> Void] {
-        guard let appcues = appcues else { return [] }
+    private func processFirstAction() {
+        guard !isProcessing, let appcues = appcues else { return }
 
-        return actionModels.compactMap { actionModel in
-            if let actionInstance = actions[actionModel.type]?.init(config: actionModel.config) {
-                return { completion in
-                    actionInstance.execute(inContext: appcues, completion: completion)
+        if let actionInstance = actionQueue.first {
+            isProcessing = true
+            actionInstance.execute(inContext: appcues) {
+                DispatchQueue.main.async {
+                    // On completion, remove the action, which triggers the didSet to process the remaining action handlers.
+                    self.isProcessing = false
+                    self.actionQueue.removeFirst()
                 }
             }
-            return nil
         }
+    }
+
+    /// Enqueue an experience action instance to be executed.
+    func enqueue(actionInstances: [ExperienceAction]) {
+        actionQueue.append(contentsOf: actionInstances)
+    }
+
+    /// Enqueue an experience action data model to be executed.
+    func enqueue(actionModels: [Experience.Action]) {
+        let actionInstances = actionModels.compactMap {
+            actions[$0.type]?.init(config: $0.config)
+        }
+        enqueue(actionInstances: actionInstances)
     }
 }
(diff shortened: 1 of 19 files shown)


Sources/AppcuesKit/Presentation/Actions/ActionRegistry.swift
```diff
@@ -42,16 +47,31 @@ internal class ActionRegistry {
...
-            if let actionInstance = actions[actionModel.type]?.init(config: actionModel.config) {
-                return { completion in
-                    actionInstance.execute(inContext: appcues, completion: completion)
+        if let actionInstance = actionQueue.first {
+            isProcessing = true
+            actionInstance.execute(inContext: appcues) {
+                DispatchQueue.main.async {
+                    // On completion, remove the action, which triggers the didSet to process the remaining action handlers.
+                    self.isProcessing = false
+                    self.actionQueue.removeFirst()
```
iujames, line 59: always looking for any race condition possibility in things like this 🤔

this _seems_ safe, since if something is enqueued while isProcessing == true, it would then get picked up and ran after the removeFirst.  

thinking out loud, but I don't see a way something will get left lingering here 👍 Kotlin made this easy with built in Channels.

edit: and you tested it well 👍

Sources/AppcuesKit/Presentation/Extensions/View+Appcues.swift
```diff
@@ -10,32 +10,21 @@ import SwiftUI
...
-    private func process(_ actionHandlers: [(@escaping ActionRegistry.Completion) -> Void]) {
-        if let handler = actionHandlers.first {
-            handler {
-                DispatchQueue.main.async {
-                    // On completion, process the remaining action handlers.
-                    process(Array(actionHandlers.dropFirst()))
-                }
-            }
-        }
-    }
```
iujames, line 23: 🔪 nice

Sources/AppcuesKit/Presentation/ExperienceRendering/ExperienceStateMachine+State.swift
```diff
@@ -40,8 +40,12 @@ extension ExperienceStateMachine {
                 )
             case let (.endingStep(experience, currentIndex, _), .startStep(stepRef)):
                 return Transition.fromEndingStepToBeginningStep(experience, currentIndex, stepRef, traitComposer)
-            case (.endingExperience, .reset):
-                return Transition(toState: .idling)
+            case let (.endingExperience(experience, stepIndex, markComplete), .reset):
+                var sideEffect: SideEffect?
+                if markComplete || stepIndex == experience.stepIndices.last {
```
iujames, line 45: one minor suggestion would be to extract this little bit of logic here and shared on line 85 of ExperienceStateObserver to a single spot, if possible -- checking if an experience was completed or not.

iujames (approved):
very similar to the Android implementation - I like it a lot, very clean handling of this I think.

Sources/AppcuesKit/Presentation/Actions/ActionRegistry.swift
mmaatttt: Always a good question to raise! You've got it exactly right. Only removing from the start and adding to the end makes it simpler. And without the `isProcessing` check, adding a new action while an existing one was running would also trigger the didSet, meaning we'd be processing twice and the last one to finish would crash trying to `removeFirst` when empty.

Sources/AppcuesKit/Presentation/ExperienceRendering/ExperienceStateMachine+State.swift
mmaatttt: Updated in f31db7dd227160260b6846c71535755dfe03fc2c because you're right that should have the logic a single place, even if it's not quite as elegant as I'd hoped.

It's tricky to do elegantly because the analytics observer has a `Result<ExperienceStateMachine.State>` versus this just being an `ExperienceStateMachine.State` so I can't nicely extend either one of those to encapsulate this logic.

Basically, if I do
```swift
extension ExperienceStateMachine.State {
        var isExperienceCompleted: Bool { ... }
}
```
then I don't have access to the `State` var in the AnalyticsObserver because it's stuck in the `Result` type, so I need extensions on both.

PR #228, opened 2022-07-26
Add shake gesture detection to refresh experience previews

mmaatttt:
Went with the notification approach since coupling the ExperienceStepViewController to the ExperienceLoader seems like a bad idea.

### DIFF
--- a/Sources/AppcuesKit/Presentation/ExperienceLoader.swift
+++ b/Sources/AppcuesKit/Presentation/ExperienceLoader.swift
@@ -19,11 +19,18 @@ internal class ExperienceLoader: ExperienceLoading {
     private let config: Appcues.Config
     private let networking: Networking
     private let experienceRenderer: ExperienceRendering
+    private let notificationCenter: NotificationCenter
+
+    /// Store the experience ID loaded (only if previewing) so that it can be refreshed.
+    private var lastPreviewExperienceID: String?
 
     init(container: DIContainer) {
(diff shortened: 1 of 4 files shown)


Sources/AppcuesKit/Presentation/ExperienceLoader.swift
```diff
@@ -42,6 +49,15 @@ internal class ExperienceLoader: ExperienceLoading {
                 self?.config.logger.error("Loading experience %{public}s failed with error %{public}s", experienceID, "\(error)")
                 completion?(.failure(error))
             }
+
+            self?.lastPreviewExperienceID = published ? nil : experienceID
```
iujames, line 53: just wondering if this should also clear out some way when the preview is dismissed - i.e. only allow shake to refresh when preview is up?  thinking of this use case: preview something, dismiss it, do other stuff for 10 mins, trigger a shake gesture - preview comes back.  This is a pretty far edge case of concern though.

iujames (approved):

Sources/AppcuesKit/Presentation/ExperienceLoader.swift
mmaatttt: The shake gesture is only detected in the `ExperienceStepViewController` so once that's out the of the responder chain, the shake detection is gone too. If different experience has loaded then this will have reset. So it's not the most direct mechanism, but it should be safe.

PR #229, opened 2022-08-02
Add ability to verify deeplink configuration from within the debugger

mmaatttt:
This was actually quite simple: for the first check we inspect the `Info.plist`, then we call a new deeplink url that doesn't do anything other than tell the debugger that the link was received. There's a unique token for each call just in case there's any weirdness with multiple SDK instances or something like that.

Video here: https://a.cl.ly/12uLZwxl

### DIFF
--- a/Sources/AppcuesKit/AppcuesKit.docc/URLSchemeConfiguring.md
+++ b/Sources/AppcuesKit/AppcuesKit.docc/URLSchemeConfiguring.md
@@ -79,6 +79,17 @@ var body: some Scene {
 
 ## Verifying the Custom URL Scheme
 
-Test that the URL scheme handling is set up correctly by navigating to `appcues-APPCUES_APPLICATION_ID://sdk/debugger` in your browser on the device with the app installed.
+The Appcues debugger allows you to easily validate that the Appcues deeplink is properly configured.
+
+1. Launch the debugger in your app with a call to ``Appcues/debug()``.
+2. Expand the debugger by tapping the floating button.
+3. Tap the "Appcues Deeplink Configured" row to verify the status. If a checkmark appears, the Appcues deeplink is properly configured. 
+
+### Troubleshooting
+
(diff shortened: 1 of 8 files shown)


Sources/AppcuesKit/AppcuesKit.docc/URLSchemeConfiguring.md
```diff
@@ -79,6 +79,17 @@ var body: some Scene {
 
 ## Verifying the Custom URL Scheme
 
-Test that the URL scheme handling is set up correctly by navigating to `appcues-APPCUES_APPLICATION_ID://sdk/debugger` in your browser on the device with the app installed.
+The Appcues debugger allows you to easily validate that the Appcues deeplink is properly configured.
+
+1. Trigger the deeplink in your app with a call to ``Appcues/debug()``.
```
iujames, line 84: `Trigger the deeplink` --> `Launch the debugger`

iujames (approved):
nice! kind of a chicken/egg thing for some, as the deeplink would be used to launch the debugger often? but could help others.

can we make sure it gets ticketed/cloned for Android please?

Sources/AppcuesKit/AppcuesKit.docc/URLSchemeConfiguring.md
mmaatttt: Whoops, thanks 🤦‍♂️

mmaatttt:
> nice! kind of a chicken/egg thing for some, as the deeplink would be used to launch the debugger often? but could help others.
True. It's really only helpful for the developer first doing the implementation, which is why I made the docs launch the debugger programmatically (typo aside). Also might provide peace of mind to non-technical users who may not understand that they used the deep link to get to the debugger?

> can we make sure it gets ticketed/cloned for Android please?
Yep: https://app.shortcut.com/appcues/story/39787

PR #230, opened 2022-08-03
Remove request review action

mmaatttt:
This will be split into a new package of actions


iujames (approved):
hey its not ⚰️ ! Just (moving-emoji)

PR #231, opened 2022-08-08
Add @appcues/background-content trait

mmaatttt:
This PR introduces `ExperienceTraitLevel` which allows a trait instance to know it's context in the data model and modify its functionality accordingly. The new `@appcues/background-content` trait takes advantage of this to apply background content to either the container or the step.

Docs have been updated to try make all this clear.


iujames (approved):
very nice, awesome we got this in for v1! kudos for leading the charge to make it happen

PR #232, opened 2022-08-09
Deep link spelling

mmaatttt:
A truly vital PR...

I created a lint rule that will remind us to be consistent in the future:

Screen Shot 2022-08-09 at 11 20 54 AM

### DIFF
--- a/Examples/DeveloperCocoapodsExample/CocoapodsExample/DeeplinkNavigator.swift
+++ b/Examples/DeveloperCocoapodsExample/CocoapodsExample/DeeplinkNavigator.swift
@@ -1,5 +1,5 @@
 //
-//  DeeplinkNavigator.swift
+//  DeepLinkNavigator.swift
 //  AppcuesCocoapodsExample
 //
 //  Created by Matt on 2022-03-04.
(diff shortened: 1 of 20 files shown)


iujames (approved):
WOW

Examples/DeveloperCocoapodsExample/CocoapodsExample/DeeplinkNavigator.swift
```diff
@@ -1,5 +1,5 @@
 //
-//  DeeplinkNavigator.swift
+//  DeepLinkNavigator.swift
```
iujames, line 2: @mmaatttt I suppose this filename should change to match

mmaatttt: Yes, I was sure I did that but I guess not 🙃 Thanks for catching it

iujames: only because I hit the same issue where git didn't pickup my rename, sigh

PR #235, opened 2022-08-11
Update public interface docs

iujames:
I was doing some copy/paste/editing from docs here to the Flutter plugin to finish that out, and noticed a few small tweaks I wanted to propose here.


mmaatttt (approved):
Good changes!

PR #236, opened 2022-08-17
Improve support for dynamic content sizing with background content

iujames:
Fix up an issue noticed while testing out embedded content ideas - this should prevent any container sizing from being based upon the preferred size of a background content ViewController (which has height zero).

### DIFF
--- a/Sources/AppcuesKit/Presentation/Extensions/UIViewController+Embed.swift
+++ b/Sources/AppcuesKit/Presentation/Extensions/UIViewController+Embed.swift
@@ -19,7 +19,7 @@ extension UIViewController {
     func unembedChildViewController(_ childVC: UIViewController) {
         guard childVC.parent == self else { return }
         childVC.willMove(toParent: nil)
-        childVC.removeFromParent()
         childVC.view.removeFromSuperview()
+        childVC.removeFromParent()
     }
 }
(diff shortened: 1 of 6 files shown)


Sources/AppcuesKit/Presentation/Extensions/UIViewController+Embed.swift
```diff
@@ -19,7 +19,7 @@ extension UIViewController {
     func unembedChildViewController(_ childVC: UIViewController) {
         guard childVC.parent == self else { return }
         childVC.willMove(toParent: nil)
-        childVC.removeFromParent()
         childVC.view.removeFromSuperview()
+        childVC.removeFromParent()
```
iujames, line 23: To get `.isMovingFromParent` to work correctly on disappear, the `removeFromParent()` needs to be called after the `view.RemoveFromSuperview()`.  Not breaking anything currently, I don't think, but would make life easier in the future to have organized this way so that we can properly track dismissal/close analytics on any non-modal content.

mmaatttt: I'm not sure why we had it in that order since Apple's code examples match what you've changed it to, so no concerns about this breaking anything: https://developer.apple.com/library/archive/featuredarticles/ViewControllerPGforiPhoneOS/ImplementingaContainerViewController.html

PR #237, opened 2022-08-19
for discussion: embedded experiences

iujames:
Larger one here but hopefully the 4 commits are logically structured in sequence:

1. Refactoring the ExperienceRenderer to support multiple ExperienceStateMachines.  The ExperienceRender now has a primary `sharedStateMachine` and then a mapping of ID -> `ExperienceStateMachine` for any secondary non-modal experiences that are allowed to run in parallel. There can only be one Experience in the main shared machine at a time - normal modal operation, but now non-modal embeds could take advantage of concurrent rendering.  The functions to dismiss an experience or move to a different step in an experience now take an ID value to allow the renderer to know which state machine those operations should apply to.  In all cases where an Experience ID is not set or otherwise unknown, it is assumed it operates on the shared modal instance.

2. Add the experience ID to the config used in ExperienceActions - this allows actions in an Experience to communicate the proper experience to apply to, when interfacing with the ExperienceRenderer.  Specifically, this is used in the close and continue actions - where the Experience context is needed so that the action applies to the correct state machine - which could be the shared modal machine or any one of the independent non-modal machines.  The ID used here is the `experience.instanceID` (not the `experience.id`) -- this is so that hypothetically the same Experience could even be running in multiple places, but have its own state machine context to run in independently.  Using the config map for this allows us to introduce this capability with no change to the public interface of the SDK in the ExperienceAction execute.

3. Add the AppcuesEmbedTrait and support for embedded content in the AppcuesView -- this is the main UI related change.  The AppcuesView is a UIView, public in the SDK, which customer apps can include anywhere desired.  The main Appcues interface allows a registration function call to register an embed with an ID tied to a view.  The SDK then retains a weakly held collection.  The AppcuesEmbedTrait is then used in an experience that would like to target a specific embed by ID - looking up the associated AppcuesView (in TraitComposer logic) and embedding the experience content inside that view, rather than opening up a new modal like current experiences use.  There is a small change to the skippable trait to handle the differences between modal dismissal and non-modal removal, when closing an experience.

4.  Update the example app - just some example using in the SPM example to show how an embed could be used.  To demonstrate, the [example experience response here](https://github.com/appcues/appcues-mobile-experience-spec/blob/feature/sc-40023/embeds/experience-examples/embed_one.json) can be used.

To use one of our experiences in an embed, you would first remove the modal trait, then add an experience level embed trait like: 
```
{
    "type": "@appcues/embed",
    "config": {
        "embedId": "slot1",
	"animated": true,
	"style": {
	    "cornerRadius": 4
	}
    }
}
```
The key value is the `embedId` that tells the SDK which registered view to target the embed into.  The `animated` option controls whether show/hide uses a subtle animation.  The same high level styling can be used on the embed (cornerRadius, margin, padding).

Embedded experiences should really just contain a single step or a single group of steps (carousel).  Multiple step groups seems like it would be undesired here, but in theory, it could still be supported as different experience content containers would be added/removed in the same host view.

Screen Shot 2022-08-19 at 8 49 12 AM

Screen Shot 2022-08-19 at 8 48 47 AM


Sources/AppcuesKit/Presentation/Traits/TraitComposer.swift
```diff
@@ -152,3 +165,16 @@ private extension Array {
         return nil
     }
 }
+
+@available(iOS 13.0, *)
+extension Appcues {
+    func resolveEmbedViewFor(experience: Experience) -> AppcuesView? {
```
mmaatttt, line 171: As far as I can tell it doesn't really matter either way, so mostly just curious if there's an advantage to this being  `extension Appcues` vs a private method in `TraitComposer`?

Sources/AppcuesKit/Presentation/UI/AppcuesView.swift
```diff
@@ -0,0 +1,121 @@
...
+import UIKit
+
+@available(iOS 13.0, *)
+public class AppcuesView: UIView {
+
+    private var isConfigured = false        // appcues instance, embed ID, and view controller set
+    private var isAttachedToView = false    // didMoveToSuperView has been called
+    private var hasTracked = false          // has tracked the embed to check for qualified content
+
+    private var appcues: Appcues?
```
mmaatttt, line 18: I think this needs to be `weak`?

And I appreciate the explicitness of having `isConfigured: Bool`, but if `appcues` becomes weak, the `let appcues = appcues` check on L52 makes checking `isConfigured` redundant.

Sources/AppcuesKit/Presentation/UI/AppcuesView.swift
```diff
@@ -0,0 +1,121 @@
...
+        // re-purposing screen for this in the short term, a deeper topic for how to best handle for embeds
+        appcues.screen(title: embedId)
+    }
+
+    override public func didMoveToSuperview() {
+        super.didMoveToSuperview()
+
+        if subviews.isEmpty {
+            // possibly have a delegate for these embeds that would allow the host app to have more control over this?
+            isHidden = true
```
mmaatttt, line 63: Do we need to be forcing the size of `AppcuesView` to zero? `isHidden` is sufficient if embedded in a `UIStackView`, but other places would leave an empty box.

Sources/AppcuesKit/Presentation/UI/AppcuesView.swift
```diff
@@ -0,0 +1,121 @@
...
+
+        trackEmbed()
+    }
+
+    internal func embed(_ experienceController: UIViewController, margins: NSDirectionalEdgeInsets, animated: Bool) {
+        guard let viewController = viewController else { return }
+
+        heightObserver = experienceController.observe(\.preferredContentSize) { _, _ in
+            self.preferredHeightConstraint.constant = experienceController.preferredContentSize.height
+        }
```
mmaatttt, line 76: This closure may need a `[weak self]` to avoid a memory leak? Would be simple enough either way since it'd just become `self?`.

Sources/AppcuesKit/Presentation/UI/AppcuesView.swift
```diff
@@ -0,0 +1,121 @@
...
+            isHidden = true
+            viewController?.unembedChildViewController(experienceController)
+            self.experienceController = nil
+        }
+    }
+
+    internal func remove() {
+        guard let viewController = viewController else { return }
+        embedTrait?.remove(viewController: viewController, completion: nil)
+    }
```
mmaatttt, line 119: This is very clever to have skippable call this which calls the embed trait dismisser!

Sources/AppcuesKit/Presentation/Traits/Appcues/AppcuesSkippableTrait.swift
```diff
@@ -32,7 +32,12 @@ internal class AppcuesSkippableTrait: ContainerDecoratingTrait, BackdropDecorati
 
     @objc
     private func didTapBackground() {
-        containerController?.dismiss(animated: true)
+
+        if let embedViewController = containerController?.view.superview as? AppcuesView {
```
mmaatttt, line 36: Can this be `if let embedView` since it's not a VC?

Sources/AppcuesKit/Presentation/Traits/Appcues/AppcuesEmbedTrait.swift
```diff
@@ -0,0 +1,102 @@
...
+
+        containerController.view.clipsToBounds = true
+
+        containerController.view.backgroundColor = UIColor(dynamicColor: embedStyle?.backgroundColor)
+        containerController.view.layer.cornerRadius = embedStyle?.cornerRadius ?? 0
+
+        containerController.view.layer.borderColor = UIColor(dynamicColor: embedStyle?.borderColor)?.cgColor
+        containerController.view.layer.borderWidth = CGFloat(embedStyle?.borderWidth) ?? 0
+
+        // containerController.view.shadowLayer = CAShapeLayer(shadowModel: embedStyle?.shadow)
```
mmaatttt, line 65: Is it worth a helper (extension on `UIViewController`, maybe?) that applies the styles to a `UIViewController.view` since this is the same as what we've got in `DialogContainerViewController`?

Sources/AppcuesKit/Presentation/ExperienceRendering/ExperienceRenderer.swift
```diff
@@ -12,41 +12,136 @@ import UIKit
...
+            // if it is modal, use the shared state machine
+            stateMachine = sharedStateMachine
+        }
+
+        show(experience: experience, in: stateMachine, priority: priority, published: published, completion: completion)
+    }
+
+    func show(qualifiedExperiences: [Experience], priority: RenderPriority, completion: ((Result<Void, Error>) -> Void)?) {
+        let nonModal = qualifiedExperiences.filter { $0.isNonModal }
+        let modal = qualifiedExperiences.filter { !$0.isNonModal }
```
mmaatttt, line 70: I know performance isn't a big concern for a list of experiences since it's small, but you could do this:

```swift
extension Collection {
    func separate(predicate: (Iterator.Element) -> Bool) -> (matching: [Iterator.Element], notMatching: [Iterator.Element]) {
        var separated: ([Iterator.Element],[Iterator.Element]) = ([],[])
        for element in self {
            if predicate(element) {
                separated.0.append(element)
            } else {
                separated.1.append(element)
            }
        }
        return separated
    }
}

let (nonModal, modal) = qualifiedExperiences.separate { $0.isNonModal }
```

Sources/AppcuesKit/Presentation/ExperienceRendering/ExperienceRenderer.swift
```diff
@@ -12,41 +12,136 @@ import UIKit
...
+
+        // for the overall success/failure completion - we have potentially mulitple experiences being rendered from the
```
mmaatttt, line 72: Typo `mulitple`

Sources/AppcuesKit/Presentation/ExperienceRendering/ExperienceRenderer.swift
```diff
@@ -57,7 +152,7 @@ internal class ExperienceRenderer: ExperienceRendering {
         // only track analytics on published experiences (not previews)
         // and only add the observer if the state machine is idling, otherwise there's already another experience in-flight
         if published && stateMachine.state == .idling {
-            self.stateMachine.addObserver(analyticsObserver)
+            stateMachine.addObserver(ExperienceStateMachine.AnalyticsObserver(container: container))
```
mmaatttt, line 155: I think if we wanted to, we could keep the single `analyticsObserver` and use it in all the state machine instances since it only uses the data it's passed and has no state otherwise. Not sure if that'd be more or less confusing.

Sources/AppcuesKit/Presentation/ExperienceRendering/ExperienceRenderer.swift
```diff
@@ -167,3 +281,35 @@ internal class ExperienceRenderer: ExperienceRendering {
         }
     }
 }
+
+@available(iOS 13.0, *)
+private extension ExperienceRenderer {
+
+    // helper observer used to clean up the mapping of experience state machines when
+    // an experience is completed
+    class StateMachineCleanupObserver: ExperienceStateObserver {
```
mmaatttt, line 290: I quite like how the observers make this cleanup easy 💯

mmaatttt (commented):
This is sooo good! I've got a hodge podge of assorted comments, but overall I'm amazed at how quickly and nicely this came together.

Sources/AppcuesKit/Appcues.swift
```diff
@@ -257,6 +268,18 @@ public class Appcues: NSObject {
...
     }
 
+    /// Registers the given embed view to be available for targeting embedded Appcues experience content.
+    /// - Parameters:
+    ///   - view: The AppcuesView to register for hosting embedded content.
+    ///   - embedId: The unique identifier for the embedded AppcusView.
+    ///   - viewController: The UIViewController that owns the provided AppcuesView instance.
+    @available(iOS 13.0, *)
+    @objc
+    public func registerEmbed(_ view: AppcuesView, embedId: String, on viewController: UIViewController) {
```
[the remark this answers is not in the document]
iujames: So I believe the reason this is required here is that the function signature uses a type `AppcuesView` which is `@available(iOS 13.0, *)`, so the entire function has to be marked this way.  Unlike the others where they are just guarding against some other iOS 13 usage internally but do not expose a type that requires it in the function signature.

Sources/AppcuesKit/Presentation/Traits/TraitComposer.swift
```diff
@@ -36,6 +38,8 @@ internal class TraitComposer: TraitComposing {
             throw ExperienceStateMachine.ExperienceError.step(experience, stepIndex, errorMessage)
         }
 
+        let embedView = appcues?.resolveEmbedViewFor(experience: experience)
```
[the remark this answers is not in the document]
iujames: yep good call, this was an artifact of some other refactoring but belongs down there now - will move.

Sources/AppcuesKit/Presentation/Actions/ActionRegistry.swift
```diff
@@ -67,9 +67,13 @@ internal class ActionRegistry {
...
 
     /// Enqueue an experience action data model to be executed.
-    func enqueue(actionModels: [Experience.Action]) {
-        let actionInstances = actionModels.compactMap {
-            actions[$0.type]?.init(config: $0.config)
+    func enqueue(actionModels: [Experience.Action], experienceID: String?) {
+        let actionInstances = actionModels.compactMap { model -> ExperienceAction? in
+            var configMap = model.config ?? [:]
+            if let experienceID = experienceID {
+                configMap["experienceID"] = experienceID
```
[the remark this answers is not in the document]
iujames: that seems like a really good idea! will do

Sources/AppcuesKit/Presentation/Traits/TraitComposer.swift
iujames: right - I'll move into TraitComposer for simplicity 👍

Sources/AppcuesKit/Appcues.swift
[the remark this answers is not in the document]
mmaatttt: Right, that makes sense. It wouldn't be impossible to make it work silently for older versions, but that might actually be more confusing, so leaving this as is probably makes sense.

For programmatic views, the developer could decide whether to use an `AppcuesView` or not, but I don't think it'd be possible in a storyboard for an app that supports back to iOS 11. If we're really gunning for simplicity, maybe making it silently fail for older versions would be the way to go?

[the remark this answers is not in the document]
iujames: yep - so effectively, remove the `@available(iOS 13.0, *)` on `AppcuesView` and have it silently fail in spots and never register views with the SDK if < 13 ?

Sources/AppcuesKit/Presentation/Traits/Appcues/AppcuesEmbedTrait.swift
```diff
@@ -0,0 +1,102 @@
...
+        guard let stepController = stepController as? ExperienceStepViewController else { return }
+        stepController.padding = NSDirectionalEdgeInsets(
+            top: embedStyle?.paddingTop ?? 0,
+            leading: embedStyle?.paddingLeading ?? 0,
+            bottom: embedStyle?.paddingBottom ?? 0,
+            trailing: embedStyle?.paddingTrailing ?? 0)
+    }
+
+    func createWrapper(around containerController: ExperienceContainerViewController) throws -> UIViewController {
+        // should throw at this point if we don't have an embedViewController to use?
```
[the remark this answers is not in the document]
iujames: right, its `embedView` now, and I'll check here and throw as well to exit early when we know the trait cannot function.

Sources/AppcuesKit/Presentation/Traits/TraitComposer.swift
```diff
@@ -106,6 +110,13 @@ internal class TraitComposer: TraitComposing {
 
         let unwrappedPresenting = try presenting.unwrap(or: TraitError(description: "Presenting capability trait required"))
 
+        // special case - embeds - provide the trait access to the embed container view controller
+        // and vice-versa (for dismissal)
+        if let embedPresenting = unwrappedPresenting as? AppcuesEmbedTrait {
+            embedPresenting.embedView = embedView
+            embedView?.embedTrait = embedPresenting
```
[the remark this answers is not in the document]
iujames: I dont _think_ so, as the trait holds the reference to the view weakly.

Sources/AppcuesKit/Appcues.swift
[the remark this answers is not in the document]
mmaatttt: Yeah, I guess just have it act like a regular empty `UIView` as much as possible?

Sources/AppcuesKit/Presentation/Traits/TraitComposer.swift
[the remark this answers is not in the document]
mmaatttt: Right, all good 👍

Sources/AppcuesKit/Presentation/UI/AppcuesView.swift
iujames: good point! updated

Sources/AppcuesKit/Presentation/UI/AppcuesView.swift
iujames: yep, that makes sense, updated.

Sources/AppcuesKit/Presentation/ExperienceRendering/ExperienceRenderer.swift
iujames: yeah, i think you're right - trying it out.  simpler is better

Sources/AppcuesKit/Presentation/Traits/TraitComposer.swift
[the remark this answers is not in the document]
iujames: at the high level this function is effectively (1) decomposing all the traits to get the full collection (2) applying all the traits and building up the step and container controllers.  Maybe some split along those lines could work to start.

Sources/AppcuesKit/Presentation/Traits/Appcues/AppcuesEmbedTrait.swift
iujames: yeah, i think it works well on a UIView extension, since they are both just applying updates to a view.  Here it is the `view`.  The other case it is the `containerView.dialogView`.

Sources/AppcuesKit/Presentation/UI/AppcuesView.swift
iujames: setting the height constraint to zero at this point leads to broken behavior with the dynamic sizing for some reason 🤔 working through that one.

iujames: @mmaatttt yep, thanks.  I did a few experiments on it and pushed up where I've landed so far here.  The goal is to have something that is totally self-contained and dynamic sizing.  It seems like the approach in there now works - basically, have two sets of constraints - the ones used when the content is empty (zero height/width) and the ones used when the content exists and should be dynamically sized ( >= 1 height/width).  Interestingly, the height observation stuff seems totally unnecessary, as the other constraints seem to handle it.

Sources/AppcuesKit/Presentation/Traits/TraitComposer.swift
```diff
@@ -83,8 +85,20 @@ internal class TraitComposer: TraitComposing {
...
 
+        let unwrappedPresenting = try presenting.unwrap(or: TraitError(description: "Presenting capability trait required"))
+
+        // special case - embeds - provide the trait access to the embed container view controller
+        // and vice-versa (for dismissal)
+        if let embedPresenting = unwrappedPresenting as? AppcuesEmbedTrait {
+            let embedView = resolveEmbedViewFor(experience: experience)
+            embedPresenting.embedView = embedView
+            embedView?.embedTrait = embedPresenting
+        }
```
mmaatttt, line 96: `try presenting.unwrap` was previously down after `wrapperCreating`, just before the `return`. Does it need to be moved up to here? (Trying to fail early perhaps?)

I'm asking because the docs list the order traits are applied (https://appcues.github.io/appcues-ios-sdk/documentation/appcueskit/traits#Trait-Application-Sequence) and I ordered them that way to try and be as easy to understand as possible (i.e. "prepare everything and then present it"). Realistically there's not many cases where the application order should matter.

iujames: yes, it can be moved back I think. It was here only to allow the AppcuesEmbedTrait to know that the view was not found earlier, during `createWrapper` - but it would not be required to be in that order.  It can fail later in `present`.

iujames:
closing this PR for now and leaving the branch, pending decisions about path forward for v 1.1 and beyond.

PR #238, opened 2022-08-25
Add basic UI implementation for form components

mmaatttt:
This PR maps the data models and implements _most_ of the UI functionality.

Currently unimplemented or rough around the edges:
- ~multiline text, max input length~
- keyboard safe area management for the text field
- ~checkbox/radio control styling (json model updates tbd)~

These additional primitives along with their complexity are creating some strange CPU and memory usage issues within SwiftUI. The update to `View+Appcues` wrapping `AnyView` fixes those and doesn't seem to have any other side effects despite `AnyView` generally being an anti-pattern. Definitely something to keep an eye on.

Otherwise, this is actually fairly simple for what it enables. All the form state is stored in the `ExperienceStepViewModel` and so we'll be able to grab the info out of there to send the necessary form interaction events in a future PR.

`ExperienceStepViewModel.FormItem` is how I'm generalizing the interface for selecting a single vs multiple values: The enum holds a `String` or `Set<String>` and then `get`, `set`, and `contains` functions manipulate those values. This will allow us to send a comma-separated list of values for the multi-select case.

### DIFF
--- a/Sources/AppcuesKit/Presentation/Extensions/View+Appcues.swift
+++ b/Sources/AppcuesKit/Presentation/Extensions/View+Appcues.swift
@@ -110,12 +110,15 @@ extension View {
     }
 
     func applyAllAppcues(_ style: AppcuesStyle) -> some View {
-        self
-            .applyForegroundStyle(style)
-            .applyInternalLayout(style)
-            .applyBackgroundStyle(style)
-            .applyBorderStyle(style)
-            .applyExternalLayout(style)
+        // Using `AnyView` here drastically improves memory and CPU usage
+        AnyView(
+            self
+                .applyForegroundStyle(style)
+                .applyInternalLayout(style)
+                .applyBackgroundStyle(style)
+                .applyBorderStyle(style)
+                .applyExternalLayout(style)
+        )
     }
 }
 
--- a/Sources/AppcuesKit/Presentation/UI/Components/MultilineTextView.swift
+++ b/Sources/AppcuesKit/Presentation/UI/Components/MultilineTextView.swift
@@ -1,0 +1,169 @@
+//
+//  MultilineTextView.swift
+//  AppcuesKit
+//
+//  Created by Matt on 2022-09-12.
+//  Copyright © 2022 Appcues. All rights reserved.
+//
+
+import SwiftUI
+
+@available(iOS 13.0, *)
+internal struct MultilineTextView: UIViewRepresentable {
+    @Binding var text: String
+    let model: ExperienceComponent.TextInputModel
+
+    func makeCoordinator() -> Coordinator {
+        Coordinator(self)
+    }
+
+    func makeUIView(context: Context) -> UIView {
+        if (model.numberOfLines ?? 1) > 1 {
+            return makeTextView(context: context)
+        } else {
+            return makeTextField(context: context)
+        }
+    }
+
+    func updateUIView(_ uiView: UIView, context: Context) {
+        if let textView = uiView as? UITextView {
+            textView.text = text
+        } else if let textField = uiView as? UITextField {
+            textField.text = text
+        }
+    }
+
+    private func makeTextView(context: Context) -> UITextView {
+        let textView = UITextView()
+        textView.delegate = context.coordinator
+        textView.backgroundColor = .clear
+        textView.textContainer.lineFragmentPadding = 0
+        textView.textContainerInset = UIEdgeInsets(top: 16, left: 16, bottom: 16, right: 16)
+
+        let fontSize = model.textFieldStyle?.fontSize ?? UIFont.labelFontSize
+        textView.font = UIFont.matching(name: model.textFieldStyle?.fontName, size: fontSize)
+        textView.textColor = UIColor(dynamicColor: model.textFieldStyle?.foregroundColor)
+
+        textView.tintColor = UIColor(dynamicColor: model.cursorColor)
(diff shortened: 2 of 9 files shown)


Sources/AppcuesKit/Presentation/Extensions/View+Appcues.swift
```diff
@@ -110,12 +110,15 @@ extension View {
...
-            .applyExternalLayout(style)
+        // Using `AnyView` here drastically improves memory and CPU usage
+        AnyView(
+            self
+                .applyForegroundStyle(style)
+                .applyInternalLayout(style)
+                .applyBackgroundStyle(style)
+                .applyBorderStyle(style)
+                .applyExternalLayout(style)
+        )
```
iujames, line 121: curious how you found this approach - is this something others have seen as well?

Sources/AppcuesKit/Presentation/UI/Components/SelectToggleView.swift
```diff
@@ -0,0 +1,76 @@
...
+
+    @Binding var selected: Bool
+    let appearance: Appearance
+    let controlPosition: ExperienceComponent.OptionSelectModel.ControlPosition?
+    @ViewBuilder var content: Content
+
+    @ViewBuilder var control: some View {
+        switch controlPosition {
+        case .leading:
+            HStack(spacing: 0) {
```
iujames, line 26: checking understanding, so any spacing would be determined by the margin on the content item itself?

Sources/AppcuesKit/Presentation/UI/ExperienceStepViewModel.swift
```diff
@@ -20,6 +20,12 @@ internal class ExperienceStepViewModel: ObservableObject {
     private let actions: [UUID: [Experience.Action]]
     private let actionRegistry: ActionRegistry?
 
+    @Published private var formComponents: [UUID: FormItem]
+
+    var formIsComplete: Bool {
+        !formComponents.contains { !$0.value.isSatisfied }
+    }
```
iujames, line 27: assume this is coming later, but curious where this ties in to the action flow

iujames (approved):
looking good!

Sources/AppcuesKit/Presentation/Extensions/View+Appcues.swift
mmaatttt: So I was trying to trace backwards what was causing the performance issues... My hack-n-slash approach of commenting different bits of code to see what changed led me to the `.applyBackgroundStyle(style)` part in particular.
My interpretation is that all the conditionals we have to apply the styling creates a massively complex type definition that has SwiftUI somehow recomputing _something_. Type erasing with `AnyView` seems to put that to rest. I'd love to understand more and will definitely be digging more into this.

A few interesting articles:
https://nalexn.github.io/anyview-vs-group/
https://chsxf.dev/2022/08/28/5-tup-why-i-quit-using-swiftui.html

Sources/AppcuesKit/Presentation/UI/ExperienceStepViewModel.swift
mmaatttt: So I was experimenting with something in the Button component that checks `viewModel.formIsComplete` and disables the button, but that won't work if we only want to disable some buttons (depending on the action).

Sources/AppcuesKit/Presentation/UI/Components/SelectToggleView.swift
mmaatttt: Correct. Having baked-in styles that can't be changed seems like a generally bad idea.

Sources/AppcuesKit/Presentation/UI/ExperienceStepViewModel.swift
mmaatttt: I think only the continue action. The tricky part is if we want the button to look disabled? I think web flows don't disable the button (and in general that's a better UX). If we don't disable the button, then in theory the continue action (or the ExperienceRenderer method that it calls) needs to somehow check the state in the view model and not do anything if it's invalid (and maybe show an inline error message or indicator that the field is required?)

iujames: First thought is perhaps some optional validation function on ExperienceAction's that would return valid by default but allow an override hook to let any action apply validation - in a hopefully non-breaking way to the public API 🤔 .  Would also need to figure out how to package up the right set of params to that function to be able to validate against.

mmaatttt:
@iujames can you take another pass through this now that I've finished the UI implementations (minus the picker style). I could've (should've?) done the updates in a separate PR, but I think I prefer having all the context together here.

Sources/AppcuesKit/Presentation/UI/Components/MultilineTextView.swift
```diff
@@ -0,0 +1,169 @@
...
+        Coordinator(self)
+    }
+
+    func makeUIView(context: Context) -> UIView {
+        if (model.numberOfLines ?? 1) > 1 {
+            return makeTextView(context: context)
+        } else {
+            return makeTextField(context: context)
+        }
+    }
```
iujames, line 26: All I can say is wow! impressive wrapper here with conditional switching between UITextView and UITextField.  Unfortunate all the work is required, but this seems great how you pulled it off.

iujames (approved):
Looking great! nice work on all the details here

PR #241, opened 2022-09-13
Add ExperienceData model to wrap Experience object and hold form state

mmaatttt:
The impetus for this change is to make form state persist across step container presentations.

Storing the form state in the ExperienceStepViewModel obviously doesn't work anymore since that gets created/destroyed with the step group. We need something associated with the Experience itself. The solution is `ExperienceData`, a wrapper class around the `Experience` object that also includes the form state and gets used in the `ExperienceStateMachine`.

## Miscellaneous thoughts
1. By using `@dynamicMemberLookup` I was able to limit the refactoring to mostly updating `Experience->ExperienceData` rather than also having to change `experience.id->experience.model.id`.
2. Because `ExperienceData` contains an `ObservableObject`, it has to be iOS 13+, which impacts a couple other spots, but that doesn't matter much.
3. I originally hoped the `ExperienceStepViewModel` would contain the reference to the `StepState` for each step, but that didn't work because an `ObservableObject` in an `ObservableObject` doesn't propagate changes the way you might hope. Instead it's passed in as a sibling `.environmentObject`.

## Next steps

Now we have the form data in the state machine, so I think we just need to update the `AnalyticsObserver` to send the new events at the right time. Should be pretty easy. `LifecycleEvent` has access to the `FormData` to be able to construct the properties.

### DIFF
--- a/Sources/AppcuesKit/Data/Models/Experience/StepReference.swift
+++ b/Sources/AppcuesKit/Data/Models/Experience/StepReference.swift
@@ -26,7 +26,8 @@ internal enum StepReference: Equatable {
         }
     }
 
-    func resolve(experience: Experience, currentIndex: Experience.StepIndex) -> Experience.StepIndex? {
+    @available(iOS 13.0, *)
+    func resolve(experience: ExperienceData, currentIndex: Experience.StepIndex) -> Experience.StepIndex? {
         switch self {
         case .index(let index):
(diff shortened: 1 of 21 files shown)


iujames (approved):
This is looking good to me and I agree it should be pretty straightforward to carry on to analytics.  You already answered my only question about having `ExperienceStepViewModel` contain `StepState`.

PR #242, opened 2022-09-15
Add form interaction analytics events

mmaatttt:
(No rush on the review here, just wanted to get this up since I'm off tomorrow. Will loop back on Monday)

The changes in #241 set us up nicely here. The `AnalyticsObserver` tracks the new events when there's survey elements in a step.

The tricky part was handling the nested structure of the properties given we have the simplified encoder that only allows a few basic types. My solution was to extend the encoder to allow encoding of the `ExperienceData.StepState` type and then pass that type as the property value. The sweet part about this solution is the actual `StepState` object then is available in the debugger to use (and in tests) rather than a deeply nested, untyped dictionary which would be miserable to work with.

The other slightly weird thing is how I'm implementing `ExperienceData.StepState: Encodable`. Rather than `ExperienceData.FormItem` also being `Encodable`, I'm looping and encoding inline. The reason is that we need to include the `fieldId` that that's only available as the _key_ to the dictionary and not in `FormItem` itself.

Tests confirm that:
1. No form events are sent for steps without form elements (I had that wrong until I got to the tests)
2. The 3 events are properly sent when there is a form

I still need to handle the csv formatting of multiple values—it's still just comma separated at the moment.

Simulator Screen Shot - iPhone 13 Pro - 2022-09-15 at 16 48 50

### DIFF
--- a/Sources/AppcuesKit/Presentation/UI/ExperienceData.swift
+++ b/Sources/AppcuesKit/Presentation/UI/ExperienceData.swift
@@ -77,8 +77,8 @@ extension ExperienceData {
         }
     }
 
-    struct FormItem {
-        enum ValueType {
+    struct FormItem: Equatable {
+        enum ValueType: Equatable {
             case single(String)
             case multi(Set<String>)
 
@@ -178,6 +178,13 @@ extension ExperienceData.StepState: Encodable {
             try itemContainer.encode(formItem.getValue(), forKey: .value)
             try itemContainer.encode(formItem.label, forKey: .label)
         }
+    }
+}
+
+@available(iOS 13.0, *)
+extension ExperienceData.StepState: Equatable {
+    static func == (lhs: ExperienceData.StepState, rhs: ExperienceData.StepState) -> Bool {
+        lhs.formItems == rhs.formItems
     }
 }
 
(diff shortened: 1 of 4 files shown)


Sources/AppcuesKit/Presentation/ExperienceRendering/ExperienceStateObserver.swift
```diff
@@ -101,6 +101,7 @@ extension ExperienceStateMachine {
             case let .success(.renderingStep(experience, stepIndex, _, isFirst: false)):
                 trackLifecycleEvent(.stepSeen, LifecycleEvent.properties(experience, stepIndex))
             case let .success(.endingStep(experience, stepIndex, _)):
+                trackFormCompletion(experience, stepIndex)
```
iujames, line 104: this is causing me some uncertainty still, as it would submit a form even if the user chose to dismiss the flow - trying to see what the desired behavior really is.  Having to find more conditional ways to trigger this outside of just during step completion is also concerning though - might complicate the code a bit.

iujames (approved):

Sources/AppcuesKit/Presentation/ExperienceRendering/ExperienceStateObserver.swift
mmaatttt: Merging this and will update when the form completion events are triggered in a subsequent PR.

PR #243, opened 2022-09-20
Fix step_completed event being improperly triggered

mmaatttt:
Then:

https://user-images.githubusercontent.com/845681/191343048-eb96a3c1-62c9-47ca-b1b6-f05939dabb9f.mp4

Now:

https://user-images.githubusercontent.com/845681/191343129-29375418-80b7-4374-af96-c15ea5a31634.mp4

I added a test case for not sending the analytics event, but that doesn't reach all the way into the logic for how `markComplete` is set. Open to ideas there since the state machine tests don't make capturing the intermediate transitions easy.

### DIFF
--- a/Sources/AppcuesKit/Presentation/ExperienceRendering/ExperienceStateMachine+State.swift
+++ b/Sources/AppcuesKit/Presentation/ExperienceRendering/ExperienceStateMachine+State.swift
@@ -15,7 +15,7 @@ extension ExperienceStateMachine {
         case beginningExperience(ExperienceData)
         case beginningStep(ExperienceData, Experience.StepIndex, ExperiencePackage, isFirst: Bool)
         case renderingStep(ExperienceData, Experience.StepIndex, ExperiencePackage, isFirst: Bool)
-        case endingStep(ExperienceData, Experience.StepIndex, ExperiencePackage)
+        case endingStep(ExperienceData, Experience.StepIndex, ExperiencePackage, markComplete: Bool)
         case endingExperience(ExperienceData, Experience.StepIndex, markComplete: Bool)
 
         func transition(for action: Action, traitComposer: TraitComposing) -> Transition? {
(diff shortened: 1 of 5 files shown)


iujames (approved):
looks good, I don't have any clear ideas on more testing, but currently looking into similar on the Android side.  My initial idea is to set up some sort of mini integration test where I initialize a state machine rendering, then execute different actions like StartStep or EndExperience (with variants for markComplete) and then watching for the presence (or lack of) the step_completed analytic event coming into the AnalyticsTracker.  This is proving complex on Android since its all funneling through a kotlin StateFlow and it's challenging to test the individual pieces like the receiver side logic.

Sources/AppcuesKit/Presentation/ExperienceRendering/ExperienceStateMachine+State.swift
```diff
@@ -154,7 +154,8 @@ extension ExperienceStateMachine.Transition {
             sideEffect = .dismissContainer(package, continuation: .startStep(stepRef))
         }
 
-        return .init(toState: .endingStep(experience, stepIndex, package), sideEffect: sideEffect)
+        // Moving to a new step is an interaction that indicates the ending step is completed
+        return .init(toState: .endingStep(experience, stepIndex, package, markComplete: true), sideEffect: sideEffect)
```
iujames, line 158: @mmaatttt noting something I found while testing related things on the Android side - if I have a button that goes backwards, previous step - it will trigger step completion, since `markComplete` is always true during this transition 🤔

iujames: update: discussion and updated reqs captured here https://app.shortcut.com/appcues/story/41699/do-not-trigger-step-completed-when-flow-is-dismissed#activity-42263 (and in linked ios ticket)

mmaatttt: Right, makes sense. I guess combined with https://app.shortcut.com/appcues/story/41902/add-step-interaction-tracking-for-mobile-button-interactions, it should be a `step_interaction` but not `step_completed`?

iujames: yeah, that makes sense.

PR #244, opened 2022-09-23
Fix serialization of NSNumber values that are not actually booleans

iujames:
Issue noticed during testing of the Capacitor plugin, as the JS side was passing in a value of `1` that was being serialized as `true` in the resulting network call.  The reason was that it was actually NSNumber(1) and our code was checking if it could be casted to a Bool first, which it could.

This is an example of the same issue, in the native code.  I trigger an event with properties like 
```swift
        Appcues.shared.track(
            name: "event1",
            properties: [
                "myInt": NSNumber(1),
                "myBool": true
            ])
```

resulting JSON output before this change:
Screen Shot 2022-09-23 at 1 26 50 PM

resulting JSON output after this change:
Screen Shot 2022-09-23 at 1 25 58 PM

This also works as expected `"myBool": NSNumber(booleanLiteral: true)` -- output is boolean value in the JSON


mmaatttt (approved):
Good find and fix 👍

PR #245, opened 2022-09-26
Add support for minimum and maximum option selects

mmaatttt:
The tests are lined up with the rules listed in the spec PR: https://github.com/appcues/appcues-mobile-experience-spec/pull/84

This PR also orders the multi select output by the order checked.

### DIFF
--- a/Sources/AppcuesKit/Presentation/UI/ExperienceData.swift
+++ b/Sources/AppcuesKit/Presentation/UI/ExperienceData.swift
@@ -37,6 +37,31 @@ internal class ExperienceData {
 
 @available(iOS 13.0, *)
 extension ExperienceData {
+
+    enum Validator: Equatable {
+        /// The item has some value.
+        case nonEmpty
+        /// At least the specified number of values have been selected.
+        case minSelections(UInt)
+        /// At most the specified number of values have been selected.
+        case maxSelections(UInt)
+
+        @available(iOS 13.0, *)
+        func isSatisfied(value: ExperienceData.FormItem.ValueType) -> Bool {
+            switch (self, value) {
+            case (.nonEmpty, _), (.minSelections, .single):
+                return value.isSet
+            case (.maxSelections, .single):
+                // Case should never apply, but logically it's always valid
+                return true
+            case let (.minSelections(min), .multi(set, _)):
+                return set.count >= min
+            case let (.maxSelections(max), .multi(set, _)):
+                return set.count <= max
+            }
+        }
+    }
+
     class FormState {
         var steps: [UUID: StepState] = [:]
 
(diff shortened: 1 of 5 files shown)


Sources/AppcuesKit/Presentation/UI/ExperienceData.swift
```diff
@@ -37,6 +37,31 @@ internal class ExperienceData {
 
 @available(iOS 13.0, *)
 extension ExperienceData {
+
+    enum Validator: Equatable {
+        /// The item has some value.
+        case nonEmpty
```
mmaatttt, line 43: This could be replaced by the `.minSelections(1)`, but I like the clarity of `.nonEmpty` and it makes more sense for a textInput where "selections" isn't a thing.

Sources/AppcuesKit/Presentation/UI/ExperienceStepViewModel.swift
```diff
@@ -73,26 +73,9 @@ extension ExperienceComponent {
...
-                    label: model.label.text,
-                    value: .single(model.defaultValue?.first ?? ""),
-                    required: model.required ?? false)
-            case .multi:
-                components[model.id] = .init(
-                    type: "optionSelect",
-                    label: model.label.text,
-                    value: .multi(Set(model.defaultValue ?? [])),
-                    required: model.required ?? false)
-            }
```
mmaatttt, line 95: I moved this logic into the init's for `FormItem`. The downside is the tests get a bit more verbose.

Sources/AppcuesKit/Presentation/UI/ExperienceData.swift
```diff
@@ -125,17 +165,27 @@ extension ExperienceData {
...
-                // Toggle the value to be included in the set.
-                underlyingValue = .multi(existingValues.symmetricDifference([newValue]))
+            case .multi(var existingValues, let maxSelections):
+                if existingValues.contains(newValue) {
+                    existingValues = existingValues.filter { $0 != newValue }
+                } else if let maxSelections = maxSelections {
+                    if existingValues.count < maxSelections {
+                        existingValues.append(newValue)
+                    } else {
+                        // Would be selecting more than the max, so no change.
```
iujames, line 175: no issue with whats here, just wondering if we'll need to give any additional UI feedback in the future, in this case of "unresponsive selection"

mmaatttt: Possibly? Realistically we're leaving it up the the customer to use a label that makes it clear only `n` items can be selected. Otherwise it would be confusing.

Because this is effectively the view model for the control, it'd be pretty straightforward to put something in the `else` block here to trigger some UI feedback.

PR #247, opened 2022-09-29
Move form submission analytics to @appcues/submit-form action

mmaatttt:
This PR introduces `ExperienceActionQueueTransforming` which modifies the array of actions triggered by a particular interaction. Any `ExperienceAction` can conform to the protocol and then be given a chance to modify the array of actions from the Experience data model before it gets enqueued to execute. This protocol is currently internal, but could easily be made public if we wanted to allow custom `ExperienceAction` implementations to have this capability (it's super powerful, and therefore super easy to mess things up).

## How Queue Transforming works

For each action in the original queue, we check if its still in the queue (it could have been removed by a previous action) and we check if it's `ExperienceActionQueueTransforming` and then apply the transformation.

A couple edge cases to note: if the order is changed, the transforms will still be applied in the order from the original queue. And newly added actions will not have their transforms applied. If we want to change this behaviour it would get much more complicated than the current `reduce` implementation, so I figured it's not worth it since there's no known use case.

## How the `@appcues/submit-form` action works

`AppcuesSubmitFormAction` implements `ExperienceActionQueueTransforming` and uses the `ExperienceRenderer` to grab the current step state and check if the form is valid. If it is valid, the queue is returned unmodified. Otherwise, the `submit-form` action and all subsequent actions are removed from the queue.

Then when it comes time to execute the action, we publish the profile update event and the `step_interaction` event like we did previously from the `AnalyticsObserver`.

Note that if the `@appcues/submit-form` action is included on the step and there's no survey components in the step layout, an empty `step_interaction` event will still fire (basically the non-existent form has no invalid components so it allows it). This is different from the prior `AnalyticsObserver` implementation, which checked `stepState.formItems.isEmpty`, but since including the `@appcues/submit-form` action is an explicit thing, submitting an empty interaction feels like the right move.

## What's next

Need to tackle the UI work to show the invalid states to wrap up sc-41062. To accomplish this, `StepState` will be updated to have a flag that the `@appcues/submit-form` action can set from the `transformQueue` function which will let the views check how they should display.

These changes also lay the groundwork for a future conditional routing actions that depend on the form state. Also now that we've exposed a way to get the current Experience via the ExperienceRenderer, it seems like it'd be possible to create actions that do things like a collapsible accordion within a step 🤔

### DIFF
--- a/Sources/AppcuesKit/Presentation/ExperienceRendering/ExperienceRenderer.swift
+++ b/Sources/AppcuesKit/Presentation/ExperienceRendering/ExperienceRenderer.swift
@@ -14,6 +14,8 @@ internal protocol ExperienceRendering: AnyObject {
     func show(qualifiedExperiences: [Experience], priority: RenderPriority, completion: ((Result<Void, Error>) -> Void)?)
     func show(stepInCurrentExperience stepRef: StepReference, completion: (() -> Void)?)
     func dismissCurrentExperience(markComplete: Bool, completion: ((Result<Void, Error>) -> Void)?)
+    func getCurrentExperienceData() -> ExperienceData?
+    func getCurrentStepIndex() -> Experience.StepIndex?
 }
 
 @available(iOS 13.0, *)
@@ -165,5 +167,13 @@ internal class ExperienceRenderer: ExperienceRendering {
                 return false
             }
         }
+    }
+
+    func getCurrentExperienceData() -> ExperienceData? {
+        stateMachine.state.currentExperienceData
+    }
+
+    func getCurrentStepIndex() -> Experience.StepIndex? {
+        stateMachine.state.currentStepIndex
     }
 }
--- a/Sources/AppcuesKit/Presentation/Actions/ActionRegistry.swift
+++ b/Sources/AppcuesKit/Presentation/Actions/ActionRegistry.swift
@@ -63,7 +63,7 @@ internal class ActionRegistry {
 
     /// Enqueue an experience action instance to be executed.
     func enqueue(actionInstances: [ExperienceAction]) {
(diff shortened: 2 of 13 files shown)


Sources/AppcuesKit/Presentation/ExperienceRendering/ExperienceRenderer.swift
```diff
@@ -14,6 +14,8 @@ internal protocol ExperienceRendering: AnyObject {
     func show(qualifiedExperiences: [Experience], priority: RenderPriority, completion: ((Result<Void, Error>) -> Void)?)
     func show(stepInCurrentExperience stepRef: StepReference, completion: (() -> Void)?)
     func dismissCurrentExperience(markComplete: Bool, completion: ((Result<Void, Error>) -> Void)?)
+    func getCurrentExperienceData() -> ExperienceData?
+    func getCurrentStepIndex() -> Experience.StepIndex?
```
iujames, line 18: Just thinking back to previous POCs on embed experiences - I think this will still work great. It will just take the experience ID as a param and fetch the the appropriate state machine.

iujames (approved):
Looks great! matches all my understanding from slack chats.

Sources/AppcuesKit/Presentation/ExperienceRendering/ExperienceRenderer.swift
mmaatttt: Yeah, I had that in the back of my mind and it should work fine like you said 👌

mmaatttt:
This should actually be linked to https://app.shortcut.com/appcues/story/41063/ios-step-interaction-analytics-for-form-submitted

PR #248, opened 2022-10-04
Add UI for form component error states

mmaatttt:
New `errorLabel` will show when appropriate. The `@appcues/submit-form` action sets `stepState.shouldShowErrors = true` and that is used to show the error states.

I introduced `TintedTextView` which is almost the same as `AppcuesText` but allows overriding the tint color to be set to the error color. It seemed strange to modify `AppcuesText` to support this, hence the new View.

### DIFF
--- a/Sources/AppcuesKit/Presentation/UI/ExperienceData.swift
+++ b/Sources/AppcuesKit/Presentation/UI/ExperienceData.swift
@@ -81,6 +81,7 @@ extension ExperienceData {
 
     class StepState: ObservableObject {
         @Published var formItems: [UUID: FormItem]
+        @Published var shouldShowErrors = false
 
         var stepFormIsComplete: Bool {
             !formItems.contains { !$0.value.isSatisfied }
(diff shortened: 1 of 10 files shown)


Sources/AppcuesKit/Presentation/UI/ExperienceData.swift
```diff
@@ -126,7 +131,7 @@ extension ExperienceData {
                 case .single(let value):
                     return value
                 case .multi(let values, _):
-                    return values.joined(separator: ",")
+                    return values.joined(separator: "\n")
```
mmaatttt, line 134: Note I snuck in the change discussed this morning to separate values with newlines rather than commas

iujames (approved):
looks good! sounded like still some uncertainty on exactly what behavior we want for the error text showing/hiding (or existing) beneath each item.  But, this looks good to progress things and revisit that later as needed.

iujames:
oh, reminder on the `*` on required items - was not sure if that was meant to be here or later.

PR #249, opened 2022-10-04
Fix step_completed event being improperly triggered on backwards navigation

mmaatttt:
There's two reasonably simple cases to handle here: swiping backwards in a carousel and a button with a negative offset.

Carousel handling:
https://github.com/appcues/appcues-ios-sdk/blob/ca5561dfe26c19f05f417f3b54fa32769d582017/Sources/AppcuesKit/Presentation/ExperienceRendering/ExperienceStateMachine.swift#L139

Button handling:
https://github.com/appcues/appcues-ios-sdk/blob/ca5561dfe26c19f05f417f3b54fa32769d582017/Sources/AppcuesKit/Presentation/ExperienceRendering/ExperienceStateMachine%2BState.swift#L182-L184

The trickiest part of this was a gross bit of code where we would check `if result.isStepCompleted` instead of the actual value of `markComplete` in the state (see the retro below for how this ended up being gross). This check meant that navigating backwards from the last step in the experience would always mark the last step as complete even after the above fixes to the `markComplete` value.

I've added a unit test for the offset check, and the analytics tests that are part of the UI test suite verify that the `step_completed` event is no longer being triggered on the backwards swipe or negative offset button, and that the updated `markComplete` logic is correct.

## A brief retro

(More for my curiosity than anything else)

Way back in the early days of the state machine, we checked if the user was on the last step when the experience was dismissed, and if so, marked the experience as completed instead of dismissed. This was a fine solution:
https://github.com/appcues/appcues-ios-sdk/blob/a7bb5d390a4b0406fcbb985fd5a43b5980458478/Sources/AppcuesKit/UI/ExperienceStateObserver.swift#L45-L49

Then, we added the `@appcues/close` action with support for `markComplete` (https://github.com/appcues/appcues-ios-sdk/pull/168). This meant the State itself now held that flag: `case endingExperience(Experience, Experience.StepIndex, markComplete: Bool)`. The mistake was not updating the rest of the codebase to use that flag, and instead maintaining the previous check alongside the new state value: https://github.com/appcues/appcues-ios-sdk/blob/038d79487159eba6bd4fa3e425a246e1719ac759/Sources/AppcuesKit/Presentation/ExperienceRendering/ExperienceStateObserver.swift#L85-L89

The 🚩 red flag was that the `markComplete` value in the `State` wasn't always correct, and we'd basically correct it in real time with the check for last step when deciding which analytics event to send.

Since then we also added `markComplete` to the `endingStep` state with the same pattern.

This has all now been fixed so that `ExperienceRenderer.dismissCurrentExperience()` and `ExperienceStateMachine.containerDidDisappear()` now are responsible for deciding if the experience should be marked complete and the State value is now always correct instead of sometimes being wrong 😎

Lesson learned: if the state in a state machine isn't actually accurate, it'll come back to bite you.

### DIFF
--- a/Sources/AppcuesKit/Data/Models/Experience/StepReference.swift
+++ b/Sources/AppcuesKit/Data/Models/Experience/StepReference.swift
@@ -26,6 +26,15 @@ internal enum StepReference: Equatable {
         }
     }
 
+    var isNegativeOffset: Bool {
+        switch self {
+        case .offset(let offset) where offset < 0:
+            return true
+        default:
+            return false
+        }
+    }
+
     @available(iOS 13.0, *)
     func resolve(experience: ExperienceData, currentIndex: Experience.StepIndex) -> Experience.StepIndex? {
         switch self {
--- a/Sources/AppcuesKit/Presentation/ExperienceRendering/ExperienceRenderer.swift
+++ b/Sources/AppcuesKit/Presentation/ExperienceRendering/ExperienceRenderer.swift
@@ -133,6 +133,11 @@ internal class ExperienceRenderer: ExperienceRendering {
     }
 
     func dismissCurrentExperience(markComplete: Bool, completion: ((Result<Void, Error>) -> Void)?) {
+        // Force `markComplete` to be true if dismissing from the last step in the experience.
+        let currentStepIndex = getCurrentStepIndex()
+        let forceMarkComplete = currentStepIndex != nil && currentStepIndex == getCurrentExperienceData()?.stepIndices.last
+        let markComplete = markComplete || forceMarkComplete
+
         guard Thread.isMainThread else {
             DispatchQueue.main.async {
                 self.dismissCurrentExperience(markComplete: markComplete, completion: completion)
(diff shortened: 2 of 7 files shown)


iujames:
I'm trying to fully understand the bug issue - can you give an example case of when this led to an incorrect result? What is the user behavior when it would trigger invalid analytics? thanks

mmaatttt:
Yeah I think you've got it. There was no bug in the previous behaviour, just a confusing implementation. Then when I thought I'd implemented this new behaviour properly, I realized it actually wasn't working in all cases because of what I described causing a bug.

> One part still throwing me a bit is why the isExperienceCompleted got removed and related logic moved into the dismissCurrentExperience function, seemingly unrelated to the step analytics part.

The `dismissCurrentExperience` implementation has this line, where it's specifying the `Action` to take (which translates to the `State` we transition to)

```swift
stateMachine.transitionAndObserve(.endExperience(markComplete: markComplete)) { result in
```

I want the `markComplete` value to be correct there (ie from the start) rather than calculating it later on. This allows us to rely solely on the `markComplete` value everywhere else rather than checking `isExperienceCompleted`. Technically I could have left the `State.endingExperience` part the way it was since only changed `State.endingStep` would have been sufficient to fix the problem, but that would be more confusing and that sort of half-fix would leave us open to a similar issue in the future.

iujames:
ok got it.  I think as long as each usage of the `. endExperience` action is updated to match what was previously in the centralized spot, no issue. The one I was wondering about was the `containerDidDisappear` in the state machine, but that one was already handled. On the Android side, there are a few more of these, as several spots use the EndExperience action directly, rather than funneling through the ExperienceRenderer (close action and skippable trait, for example) - since we did not have such a dismissCurrentExperience function there.  So that would take some refactoring to match, but fairly straightforward.

iujames (approved):

PR #252, opened 2022-10-06
Add step_interaction analytics event for button presses

mmaatttt:
Whenever a button (or, technically any view with an action) is tapped, we want to record the `step_interaction` event with some metadata about that action.

To do this, we:
1. Look up the tapped element in the JSON model, and get its text description.
2. Try and figure out what the "primary" action is from the queue and fill out a `category` and `destination` value. To do this, the actions we care about conform to the `MetadataSettingAction` protocol. We go backwards through the action queue because that heuristic (currently) gives us the actions we care about with how they're ordered by the mobile builder. This isn't foolproof, but it's *mostly right* (entirely right for the current cases) and that's good enough.
3. Prefix the action queue with a new internal `AppcuesStepInteractionAction` that actually logs the analytics event. Adding at the start of the queue ensures the action is logged before and step changes or other related experience actions. There's an interested interplay with the `@appcues/submit-form` action since that only removes the actions following it when invalid, and so the `step_interaction` event is still logged even though the form was not submitted. I think this is ok because it's logging that the button was tapped (which is true and possibly helpful info), but we could add a special case to the submit form action to remove the step interaction action too.4. 

Note that I've got the following `interactionType` values: `Button Tapped`, `Button Activated` (for accessibility activations—I think it's interesting to have this differentiated from a tap, but we could make it the same), and `Button Long Pressed`.

I've added test cases to match each of the possible action lists from the builder.

The debugger work for survey interactions also works nicely for button interactions:

Simulator Screen Shot - iPhone 13 Pro - 2022-10-06 at 14 39 11

### DIFF
--- a/Sources/AppcuesKit/Data/Extensions/Sequence+CompactMapFirst.swift
+++ b/Sources/AppcuesKit/Data/Extensions/Sequence+CompactMapFirst.swift
@@ -1,0 +1,22 @@
+//
+//  Sequence+CompactMapFirst.swift
+//  AppcuesKit
+//
+//  Created by Matt on 2022-10-06.
+//  Copyright © 2022 Appcues. All rights reserved.
+//
+
+import Foundation
+
+extension Sequence {
+    // Functionally the same as Sequence.compactMap().first(), except returns immediately upon finding the first item.
+    func compactMapFirst<ElementOfResult>(_ transform: (Element) throws -> ElementOfResult?) rethrows -> ElementOfResult? {
+        for item in self {
+            if let result = try transform(item) {
+                return result
+            }
+        }
+
+        return nil
+    }
+}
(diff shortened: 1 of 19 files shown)


iujames (approved):
This looks solid! good cleanup on the dupe code bit. The only thing I was sort of wondering about was the `component(matching id: UUID)` part but now I see it's used on demand at the time of button tap to find the element that the action is executing on. I don't have any better way and thats probably efficient enough. On android there is a data to domain model mapping layer so maybe we could handle the association during that step.

mmaatttt:
@iujames It's interesting, because each of the `Appcues*` views has a model that has the data without having to parse the hierarchy, but the extension is on `View`, so we can't access it.

If performance is a concern, we could modify the `setupActions` function to take the name: `.setupActions(on: viewModel, for: model.id, label: model.textDescription)`. For some reason I didn't think of that until now, and it's not a bad idea. What do you think?

iujames:
@mmaatttt oh nice, yeah I think that might be a fairly simple approach.  It could be `.setupActions(on: viewModel, for: model)` maybe? since `textDescription` is already in the base `ComponentModel` protocol now.

mmaatttt:
Updated. Thanks, it's much nicer now!

iujames (approved):

PR #253, opened 2022-10-07
Sanitize event properties to known types before broadcasting outside the SDK

mmaatttt:
This PR converts `Date` objects to milliseconds and `StepState` objects to nested `[String: Any]` primitive JSON-compatible types.

The cool part is unlike the x-platform implementations (eg [React Native](https://github.com/appcues/appcues-react-native-module/blob/15d2d650d1cc812bdab26d5f512c900bbda99172/ios/AppcuesReactNative.swift)), I was able to do it without repeating the `switch` for `Dictionary` and `Array`. It's a bit tricky, and I learned a bunch (thanks to [this discussion](https://forums.swift.org/t/idea-mutatingforeach-for-collections/18442)), but the solution is quite nice.

I updated the tests to handle comparing arrays and the Int64 millisecond value. The `file: StaticString = #file, line: UInt = #line` stuff just makes it easier to understand failures in Xcode.

### React Native Example

|Now|Before|
|-|-|
|after|before|

### DIFF
--- a/Sources/AppcuesKit/Data/Analytics/AnalyticsBroadcaster.swift
+++ b/Sources/AppcuesKit/Data/Analytics/AnalyticsBroadcaster.swift
@@ -23,15 +23,48 @@ internal class AnalyticsBroadcaster: AnalyticsSubscribing {
     func track(update: TrackingUpdate) {
         guard let delegate = appcues?.analyticsDelegate else { return }
 
+        var properties = update.properties
+        properties?.values.sanitize()
+
         switch update.type {
         case let .event(name, _):
-            delegate.didTrack(analytic: .event, value: name, properties: update.properties, isInternal: update.isInternal)
+            delegate.didTrack(analytic: .event, value: name, properties: properties, isInternal: update.isInternal)
         case let .screen(title):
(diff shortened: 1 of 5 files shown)


Sources/AppcuesKit/Data/Analytics/AnalyticsBroadcaster.swift
```diff
@@ -23,15 +23,48 @@ internal class AnalyticsBroadcaster: AnalyticsSubscribing {
...
+            delegate.didTrack(analytic: .group, value: groupID, properties: properties, isInternal: update.isInternal)
         case .profile:
-            delegate.didTrack(analytic: .identify, value: storage.userID, properties: update.properties, isInternal: update.isInternal)
+            delegate.didTrack(analytic: .identify, value: storage.userID, properties: properties, isInternal: update.isInternal)
+        }
+    }
+}
+
+// Implemented an extension on MutableCollection rather than duplicating code for Dictionary<String, Any> and Array<Any>.
+private extension MutableCollection where Element == Any {
```
iujames, line 43: nice 👍

iujames (approved):

PR #255, opened 2022-10-13
Fix border radius rendering to match the corner radius

mmaatttt:
As [James pointed out](https://github.com/appcues/appcues-mobile-experience-spec/pull/95#pullrequestreview-1139902647), I didn't quite get it right:

These examples have a translucent red border to see how it overlays.

||Good|Bad|
|-|-|-|
|1px border|1-good|1-bad|
|4px border|4-good|4-bad|
|10px border|10-good|10-bad|


iujames (approved):
nice! glad you could make that match up.

PR #257, opened 2022-10-17
Update timestamp date format to Double

mmaatttt:
[James noted](https://github.com/appcues/appcues-android-sdk/pull/249#discussion_r997053037) that the Date format from https://github.com/appcues/appcues-ios-sdk/pull/253 was `Int64` instead of `Double`. `Double` works safely everywhere (the other place this is used is encoding the JSON for the platform requests), so simplifying to that.


mmaatttt:
@iujames that assumption's correct since in JS(ON) a Number is a Number, and it's up to the swift side to decide how that number gets encoded. I added some tests for NetworkClient to make that assumption explicit 👍

PR #258, opened 2022-10-17
Add support for experiments

iujames:
* decode a top level `experiments` mapping on the `QualificationResponse`
* add support for an optional `experiment_id` on `Experience`
* modify `ExperienceRenderer` to support experiment logic, if there is a matching experiment for an `Experience` being rendered:
    * if group is `control` - block rendering the experience and move on to next priority, if any
    * if not (group `exposed`) - continue rendering the experience as usual
    * in both cases, trigger an `appcues:experiment_entered` analytic for the user, with the group and experiment ID

update: to make the `ExperienceRenderer` portion more compact and flexible - it now operates on `ExperienceData` objects passed in, rather than `Experience`.  `ExperienceData` can then expand to contain more metadata about the `Experience`, such as the `Experiment?` added for this use case.

### DIFF
--- a/Sources/AppcuesKit/Data/Analytics/AnalyticsTracker.swift
+++ b/Sources/AppcuesKit/Data/Analytics/AnalyticsTracker.swift
@@ -94,7 +94,7 @@ internal class AnalyticsTracker: AnalyticsTracking, AnalyticsSubscribing {
             case .success(let qualifyResponse):
                 if #available(iOS 13.0, *), let experienceRenderer = self?.container?.resolve(ExperienceRendering.self) {
                     let experiments = qualifyResponse.experiments ?? [:]
-                    let qualifiedExperienceData = qualifyResponse.experiences.map {
+                    let qualifiedExperienceData: [ExperienceData] = qualifyResponse.experiences.map {
                         var experiment: Experiment?
                         if let experimentID = $0.experimentID {
                             experiment = experiments[experimentID]
--- a/Sources/AppcuesKit/Data/Analytics/AnalyticsTracker.swift
+++ b/Sources/AppcuesKit/Data/Analytics/AnalyticsTracker.swift
@@ -96,6 +96,7 @@ internal class AnalyticsTracker: AnalyticsTracking, AnalyticsSubscribing {
                     experienceRenderer.show(
                         qualifiedExperiences: qualifyResponse.experiences,
                         priority: qualifyResponse.renderPriority,
+                        experiments: qualifyResponse.experiments ?? [:],
                         completion: nil)
                 } else {
                     self?.config.logger.info("iOS 13 or above is required to render an Appcues experience")
--- a/Sources/AppcuesKit/Data/Analytics/AnalyticsTracker.swift
+++ b/Sources/AppcuesKit/Data/Analytics/AnalyticsTracker.swift
@@ -93,11 +93,15 @@ internal class AnalyticsTracker: AnalyticsTracking, AnalyticsSubscribing {
             switch result {
             case .success(let qualifyResponse):
                 if #available(iOS 13.0, *), let experienceRenderer = self?.container?.resolve(ExperienceRendering.self) {
-                    experienceRenderer.show(
-                        qualifiedExperiences: qualifyResponse.experiences,
-                        priority: qualifyResponse.renderPriority,
-                        experiments: qualifyResponse.experiments ?? [:],
-                        completion: nil)
+                    let experiments = qualifyResponse.experiments ?? [:]
+                    let qualifiedExperienceData = qualifyResponse.experiences.map {
+                        var experiment: Experiment?
+                        if let experimentID = $0.experimentID {
+                            experiment = experiments[experimentID]
+                        }
+                        return ExperienceData($0, priority: qualifyResponse.renderPriority, published: true, experiment: experiment)
+                    }
+                    experienceRenderer.show(qualifiedExperiences: qualifiedExperienceData, completion: nil)
                 } else {
                     self?.config.logger.info("iOS 13 or above is required to render an Appcues experience")
                 }
--- a/Sources/AppcuesKit/Data/Analytics/AnalyticsTracker.swift
+++ b/Sources/AppcuesKit/Data/Analytics/AnalyticsTracker.swift
@@ -92,22 +92,29 @@ internal class AnalyticsTracker: AnalyticsTracking, AnalyticsSubscribing {
         activityProcessor.process(activity) { [weak self] result in
             switch result {
             case .success(let qualifyResponse):
-                if #available(iOS 13.0, *), let experienceRenderer = self?.container?.resolve(ExperienceRendering.self) {
-                    let experiments = qualifyResponse.experiments ?? []
-                    let qualifiedExperienceData: [ExperienceData] = qualifyResponse.experiences.map {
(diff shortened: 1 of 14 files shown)


Sources/AppcuesKit/Presentation/ExperienceRendering/ExperienceRenderer.swift
```diff
@@ -10,45 +10,84 @@ import UIKit
...
+}
+
 @available(iOS 13.0, *)
 internal class ExperienceRenderer: ExperienceRendering {
 
     private let stateMachine: ExperienceStateMachine
     private let analyticsObserver: ExperienceStateMachine.AnalyticsObserver
     private weak var appcues: Appcues?
     private let config: Appcues.Config
+    private let analyticsPublisher: AnalyticsPublishing
```
iujames, line 39: @mmaatttt this adds a new line in our dependency graph, and one crossing from the presentation side to the analytics side.  I don't love it, but didn't have any better ideas to trigger this event at the spot needed here during `show(...)`.

Sources/AppcuesKit/Data/Models/Experiment.swift
```diff
@@ -0,0 +1,13 @@
...
+//  AppcuesKit
+//
+//  Created by James Ellis on 10/17/22.
+//  Copyright © 2022 Appcues. All rights reserved.
+//
+
+import Foundation
+
+internal struct Experiment: Decodable {
+    let group: String
```
iujames, line 12: fyi, updated this to just allow any String here, not requiring "control" or "exposed" enum types - my hope is to possibly add a little future-proofing for any multivariate experiments or different naming conventions in the future.  It still checks for `"control"` explicitly in the `ExperienceRenderer` to know if it should block the experience from rending, as a special case.

Sources/AppcuesKit/Presentation/ExperienceRendering/ExperienceRenderer.swift
iujames: removed in latest refactor.

Sources/AppcuesKit/Data/Analytics/AnalyticsTracker.swift
```diff
@@ -93,10 +93,15 @@ internal class AnalyticsTracker: AnalyticsTracking, AnalyticsSubscribing {
...
-                        completion: nil)
+                    let experiments = qualifyResponse.experiments ?? [:]
+                    let qualifiedExperienceData = qualifyResponse.experiences.map {
+                        var experiment: Experiment?
+                        if let experimentID = $0.experimentID {
+                            experiment = experiments[experimentID]
+                        }
+                        return ExperienceData($0, priority: qualifyResponse.renderPriority, published: true, experiment: experiment)
+                    }
+                    experienceRenderer.show(qualifiedExperiences: qualifiedExperienceData, completion: nil)
```
iujames, line 104: @mmaatttt now pre-processing the experiments here and bundling into `ExperienceData` to send into the `ExperienceRenderer` per our convo...

Sources/AppcuesKit/Data/Models/Experiment.swift
```diff
@@ -0,0 +1,32 @@
...
+              let analyticsPublisher = appcues?.container.resolve(AnalyticsPublishing.self) else { return }
+
+        analyticsPublisher.publish(TrackingUpdate(
+            type: .event(name: "appcues:experiment_entered", interactive: false),
+            properties: [
+                "experimentId": experimentID,
+                "group": group
+            ],
+            isInternal: true))
+    }
```
iujames, line 31: these are now used during the `ExperienceRenderer.show(...)` processing, encapsulating usage of `AnalyticsPublishing` here 🤷‍♂️

this has to be two-phase - `shouldExecute` then `track` - explained in the comments in usage spot in `ExperienceRenderer.show(...)`

Sources/AppcuesKit/Presentation/ExperienceRendering/ExperienceRenderer.swift
```diff
@@ -36,33 +40,46 @@ internal class ExperienceRenderer: ExperienceRendering {
...
                 case .failure(let error):
                     completion?(.failure(error))
                 }
             }
             return
         }
 
+        // if we get here, either we did not have an experiment, or it is active and did not exit early (not control group).
+        // if an active experiment does exist, it should now track the experiment_entered analytic
+        experience.experiment?.track(appcues: appcues, experimentID: experience.experimentID)
```
iujames, line 74: this case was not tracked above on line 54, to avoid double tracking in the case of recursive call on line 64 during `dismissCurrentExperience`

Sources/AppcuesKit/Presentation/ExperienceRendering/ExperienceRenderer.swift
```diff
@@ -36,33 +40,46 @@ internal class ExperienceRenderer: ExperienceRendering {
...
         guard Thread.isMainThread else {
             DispatchQueue.main.async {
-                self.show(experience: experience, priority: priority, published: published, completion: completion)
+                self.show(experience: experience, completion: completion)
             }
             return
         }
 
-        if priority == .normal && stateMachine.state != .idling {
+        if !(experience.experiment?.shouldExecute() ?? true) {
```
mmaatttt, line 51: Nit, but I think a `guard` here is makes it immediately clearer that we have to return early and makes it slightly easier than thinking about `if !shouldExecute`.

Sources/AppcuesKit/Data/Models/Experiment.swift
mmaatttt: It does feel a bit off having the tracking in a model struct just to avoid the dependency graph change 🤷‍♂️

There's already a dependency from `AnalyticsObserver` to `AnalyticsTracking` so I could just move that line over to start from `ExperienceRendering`. I think doing this and then tracking from the `ExperienceRender` might be preferable for me than having the tracking in the model, but final call is yours.

Sources/AppcuesKit/Data/Analytics/AnalyticsTracker.swift
```diff
@@ -93,10 +93,15 @@ internal class AnalyticsTracker: AnalyticsTracking, AnalyticsSubscribing {
...
-                        priority: qualifyResponse.renderPriority,
-                        completion: nil)
+                    let experiments = qualifyResponse.experiments ?? []
+                    let qualifiedExperienceData: [ExperienceData] = qualifyResponse.experiences.map {
+                        var experiment: Experiment?
+                        if let experimentID = $0.experimentID {
+                            experiment = experiments.first { $0.experimentID == experimentID }
+                        }
+                        return ExperienceData($0, priority: qualifyResponse.renderPriority, published: true, experiment: experiment)
+                    }
```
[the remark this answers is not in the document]
iujames: good call 👍

Sources/AppcuesKit/Data/Models/Experiment.swift
iujames: I agree, updated - thanks.

mmaatttt: Updated the graph:
iOS SDK Dependency Graph

iujames: thank you!

PR #260, opened 2022-10-19
Update CI config to use XCode 14

iujames:


.circleci/config.yml
```diff
@@ -74,7 +74,7 @@ commands:
 # -------------------------
```
mmaatttt, line 74: A bit worrying that it took 8 minutes to install the mint packages. It makes sense that a new Xcode might trigger that, but then it didn't save to the cache the newly installed files. We might need to bump `3-mint` to `4-mint` (L62/68 above) to get that working again.

https://app.circleci.com/pipelines/github/appcues/appcues-ios-sdk/1115/workflows/18269238-5df6-4ed3-9fea-f38ea904ac85/jobs/953

mmaatttt (approved):

.circleci/config.yml
iujames: yep, will do 👍

iujames: well, that [first try](https://app.circleci.com/pipelines/github/appcues/appcues-ios-sdk/1116/workflows/216cd37d-9bbf-4ae8-a186-be8f0c3e5148/jobs/954) failed, hmn. 
```
Creating cache archive...
Warning: could not archive /usr/local/lib/mint - Not found
```

iujames: looks like maybe `~/.mint` is the new spot. will try that.

iujames: This still doesn't seem fully correct. The Mint stuff seems resolved. I noticed the gem dependencies were also not seeming to cache correctly and updated that key.

Now, it generates this new key from the Gemfile.lock, on restore first time - does not find it (expected)
```
No cache is found for key: 2-gems-b08gPPAD2tR9QH0gxt0VmjkyRBJEYmhV34m8f5CmOaA=
```

Then it saves it to a different key
```
Stored Cache to 2-gems-5sx39qfcZrMQx3zhxoR2Ygh2xsegEAYgzyBbZwdycs8=
```

Then next time around - another miss.  Something in the restore is updating the lock file apparently. I don't know much about how that works. https://app.circleci.com/pipelines/github/appcues/appcues-ios-sdk/1118/workflows/f2dbb697-3800-4660-9fd0-6153ff982b7a/jobs/957

mmaatttt: We might want `bundle install --deployment`? That sets `frozen` which disallows lockfile changes and `--path vendor/bundle` which is what we already have there.

https://bundler.io/v2.3/man/bundle-config.1.html

PR #261, opened 2022-10-20
Add SdkMetrics to capture timings around experience rendering

iujames:
This stacks on #258 , as it builds upon the updates to `ExperienceData` and the usage in `ExperienceRenderer`.

The approach here is a static registry of metrics by request ID - in `SdkMetrics`.  This is where other spots can track key points in the qualification/rendering lifecycle.
* the moment the analytic comes in to the sdk
* when it is sent out in the network request
* when it returns from the network
* when it starts presentation in the UI

Then, when the `experience_started` analytic is trigged on initial render, we can capture and send the new `_sdkMetrics` node in the attributes, for reporting usage.

A few approaches are used to keep this metrics collection from bloating in size at runtime
* only adds entries for things that might potentially trigger qualification - analytics sent in that are "interactive" / processed immediately (i.e. the customer app screens and events).  Does not capture entries for non-qualifying background/internal events like flow analytics
* remove items when a response comes back with no qualified experiences (common case) - since we know we'll never be rendering any thing there
* does not do any tracking for non-traditional experience rendering: ex. deeplinks, show(id), flow triggering another flow
* clean up other items that do have qualification as soon as the render metrics are sent out
* clean up all items on app backgrounding, as a safety net, since no longer current/relevant

end result gets us data like this example
Screen Shot 2022-10-20 at 3 36 47 PM
(only 9ms to process a response, parse JSON, and get into render start!! 🎉)

also updated to show in the debugger for this event, in a section at the bottom
Screen Shot 2022-10-20 at 3 28 49 PM

### DIFF
--- a/Sources/AppcuesKit/Data/Analytics/AnalyticsTracker.swift
+++ b/Sources/AppcuesKit/Data/Analytics/AnalyticsTracker.swift
@@ -46,14 +46,14 @@ internal class AnalyticsTracker: AnalyticsTracking, AnalyticsSubscribing {
             syncQueue.sync {
                 flushWorkItem?.cancel()
                 pendingActivity.append(activity)
-                flushPendingActivity()
+                flushPendingActivity(trackedAt: update.timestamp) // immediately flush, eligible for qualification, so track timing
             }
 
         case .flushThenSend:
             syncQueue.sync {
                 flushWorkItem?.cancel()
-                flushPendingActivity()
-                flush(activity)
+                flushPendingActivity() // no timing tracking on pending background activity
+                flush(activity, trackedAt: update.timestamp) // immediately flush, eligible for qualification, so track timing
             }
 
         case .queue:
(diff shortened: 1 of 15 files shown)


Sources/AppcuesKit/Data/Analytics/AnalyticsTracker.swift
```diff
@@ -80,33 +80,46 @@ internal class AnalyticsTracker: AnalyticsTracking, AnalyticsSubscribing {
...
-                        return ExperienceData($0, priority: qualifyResponse.renderPriority, published: true, experiment: experiment)
+                        experienceRenderer.show(qualifiedExperiences: qualifiedExperienceData, completion: nil)
+                    } else {
+                        self?.config.logger.info("iOS 13 or above is required to render an Appcues experience")
+                        SdkMetrics.remove(activity.requestID)
                     }
-                    experienceRenderer.show(qualifiedExperiences: qualifiedExperienceData, completion: nil)
                 } else {
-                    self?.config.logger.info("iOS 13 or above is required to render an Appcues experience")
+                    // commong case, nothing qualified - we know there was nothing to render, so just remove tracking
```
mmaatttt, line 117: typo `commong`

mmaatttt (approved):
Looks good! Keeps it nice and simple

PR #262, opened 2022-10-20
Add support for custom attribute name to store survey question results

iujames:
The main change here is simple, and mostly isolated to just the `ExperienceData` file - where a new `attributeName` is handled, and used in the profile update if present. It adds a second profile attribute, in addition to the default update.

The other changes in the handling of the `.profile` tracking type are from noticing that we were not actually batching the form submit analytics together properly, previously. In the existing code, it would send out three separate requests: a `step_interaction` for the button tap (which had a flushed triggered by next item), a `profile_update` for the attribute which was set to always send immediately, and another `step_interaction` (and maybe more) for the form submit.

These can actually all be grouped together in this case, by allowing the `.profile` item to be considered non-interactive, and batched with the other flow analytics.  This results in the final output of a single call to the API like the image below. This is probably not critical, but I propose this change just to maximize our usage of activity batching in this common scenario.
Screen Shot 2022-10-20 at 4 49 43 PM

### DIFF
--- a/Sources/AppcuesKit/Data/Analytics/AnalyticsTracker.swift
+++ b/Sources/AppcuesKit/Data/Analytics/AnalyticsTracker.swift
@@ -141,23 +141,21 @@ extension Activity {
     }
 
     mutating func append(_ activity: Activity) {
-        // the only thing we support merging here is additional events.
-        // if the user or group, or any associated user or group properties are updated,
-        // those cause any pending activity to be flushed, then the user or group
-        // activity immediately flushed as an individual item - there is no concept
-        // of merging content across user or group updates.  The reason is that the
-        // activity prior to that update needs to be associated to the correct user/group
-        // and the activity after that update needs to be associated with the new user/group
+        // this will merge any additional events or profile updates.
+        // we do not support any update to the user ID or group ID here, as a change
+        // in those values would trigger an immediate flush of pending items then the new update
+        // sent. The reason is that the activity prior to that update needs to be associated to
+        // the previous user/group and the activity after that update needs to be associated with
+        // the new user/group
         if let newEvents = activity.events {
             let existingEvents = events ?? []
             events = existingEvents + newEvents
+        }
 
-            // additional events can also cause autoproperty updates, merge those in
-            newEvents.forEach {
-                if let autoProps = $0.autoProperties {
-                    profileUpdate = (profileUpdate ?? [:]).merging(autoProps) { _, new in new }
-                }
-            }
+        // merge in any updated auto props from events or other pending profile updates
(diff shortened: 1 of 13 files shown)


Sources/AppcuesKit/Data/Analytics/AnalyticsTracker.swift
```diff
@@ -141,23 +141,21 @@ extension Activity {
...
-            // additional events can also cause autoproperty updates, merge those in
-            newEvents.forEach {
-                if let autoProps = $0.autoProperties {
-                    profileUpdate = (profileUpdate ?? [:]).merging(autoProps) { _, new in new }
-                }
-            }
+        // merge in any updated auto props from events or other pending profile updates
+        if let newProfileUpdate = activity.profileUpdate {
+            let existingProfileUpdate = profileUpdate ?? [:]
+            profileUpdate = existingProfileUpdate.merging(newProfileUpdate) { _, new in new }
```
iujames, line 158: I believe this to actually be safe, rewriting some of the previous comments above, and matching Android more closely.  As long as we are careful to never allow the actual user ID to change in the middle of a batch, we can combine the properties into a merged set.

Sources/AppcuesKit/Data/Analytics/AutoPropertyDecorator.swift
```diff
@@ -157,9 +157,3 @@ extension TrackingUpdate {
         }
     }
 }
-
-extension Event {
-    var autoProperties: [String: Any]? {
-        return attributes?["_identity"] as? [String: Any]
-    }
-}
```
iujames, line 165: now dead code

Sources/AppcuesKit/Data/Analytics/AnalyticsTracker.swift
mmaatttt: Any worries about queuing the profile update impacting future qualifications or personalizations?

mmaatttt (approved):

Sources/AppcuesKit/Data/Analytics/AnalyticsTracker.swift
iujames: I don't think so, since any qualification eligible analytics would cause an immediate flush of anything in the queue as well along with it - so it should always have the most up to date profile information at qualification time (or personalization on a qualified subsequent flow). But, if you think I'm missing something there please let me know.

mmaatttt: Right. I was overlooking the fact the we'd need a flushing event. I guess technically an app could `Appcues.show()` and then have stale personalization data, but not worried about that.

PR #263, opened 2022-10-21
Update sample app build/deploy for XCode 14

iujames:
Discovered a couple issues specific to building and deploying the Cocoapods example using an updated XCode 14 build.

1. [Xcode 14 needs selected Development Team for Pod Bundles](https://stackoverflow.com/questions/72561696/xcode-14-needs-selected-development-team-for-pod-bundles) - resolved this one using the solution here https://stackoverflow.com/a/73839370

2. [Xcode 14: Could not find transporter error](https://github.com/fastlane/fastlane/issues/20664) - resolved this one by updating the fastlane version to 2.210.1

confirmed I could then get a [deployment](https://app.circleci.com/pipelines/github/appcues/appcues-ios-sdk/1142/workflows/948bd315-c416-488c-a2e0-9ce1a1b1aae4/jobs/983)


Gemfile.lock
```diff
@@ -116,7 +116,7 @@ GEM
     faraday_middleware (1.2.0)
       faraday (~> 1.0)
     fastimage (2.2.6)
-    fastlane (2.205.2)
+    fastlane (2.210.1)
```
iujames, line 119: the actual change of note here

mmaatttt: Shouldn't the version be specified in the actual Gemfile? Right now it's just `gem "fastlane"`

mmaatttt (approved):

Gemfile.lock
iujames: It can be, if we want to fix a specific version there. Right now we just get latest when running `bundle update fastlane` and set in the lockfile. Seems worth having explicit in the Gemfile to me though. Will update.

PR #264, opened 2022-10-24
Fix crash around modifying activity set

mmaatttt:
`ActivityProcessor` maintains a `Set` of in-flight request IDs. This set needs to be accessed in a thread-safe manner, so a serial queue is used. However there was one spot where the set was modified outside the safety of the serial queue which could cause a crash if the set was simultaneously being modified from within the queue.


iujames (approved):
Nice find!

iujames:
noticed this is targeting `main` - should it branch off the `1.0.0` tag in a hotfix branch instead? or do you plan to cherry pick?

mmaatttt:
Closing this in favour of the the `release` branch approach

PR #269, opened 2022-10-24
Fix logic that links experiments and experiences in response

iujames:
Once we had a real test response to try out, discovered a misunderstanding in the requirements around how the response would look. We need to link the experiment and experience as shown here - matching on the experience ID inside of the experiments array. There is no `experiment_id` in the `experience` response object.

Screen Shot 2022-10-24 at 2 56 24 PM


mmaatttt (approved):
This JSON structure does make a little bit more sense 🙂

PR #271, opened 2022-10-26
Add a config option to pass additional auto-properties

iujames:
Planning to use from the x-plat wrappers to supply a bit more context in our events.


mmaatttt (approved):
Nice and simple!

PR #273, opened 2022-10-28
Support auto downscaling on Text when available width is too narrow

iujames:
This is an approach to use the [`.minimumScaleFactor`](https://developer.apple.com/documentation/swiftui/view/minimumscalefactor(_:)) modifier to improve the situation noted in the ticket (clipping) - likely more common with side by side content.

An example to illustrate - uses 4 side-by-side text elements, all of which have too large of a font size to fit as normally expected.

* "Hello World" - 30pt
* "X" - 150pt
* "Somereallylongwordhere" - 30pt
* "👋🏼" - 200pt  <-- the primary issue in ticket

| After | Before |
| --- | --- |
|  Screen Shot 2022-10-28 at 4 42 20 PM  | Screen Shot 2022-10-28 at 4 42 56 PM  |

### DIFF
--- a/Sources/AppcuesKit/Presentation/Extensions/View+Appcues.swift
+++ b/Sources/AppcuesKit/Presentation/Extensions/View+Appcues.swift
@@ -129,7 +129,7 @@ extension View {
 
 @available(iOS 13.0, *)
 extension Text {
-    func applyTextStyle(_ style: AppcuesStyle) -> some View {
+    func applyTextStyle(_ style: AppcuesStyle, text: String) -> some View {
         self
             .ifLet(style.letterSpacing) { view, val in
                 view.kerning(val)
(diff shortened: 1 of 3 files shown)


Sources/AppcuesKit/Presentation/Extensions/View+Appcues.swift
```diff
@@ -129,13 +129,21 @@ extension View {
...
                 view.multilineTextAlignment(val)
             }
+            // The following two lines allow text to scale down in cases where it is too
+            // large to fit in the bounds available, and otherwise unnatural clipping
+            // or wrapping would occur. For example, a large emoji in a side-by-side layout.
+            // SwiftUI requires a line limit to be set for the minimumScaleFactor to be applied,
+            // and it must be set to 1 line in the case of a single large character for scaling
+            // to be applied.
+            .lineLimit(text.count == 1 ? 1 : Int.max)
+            .minimumScaleFactor(text.count == 1 ? 0.1 : 0.5)
```
iujames, line 147: Guesswork here - seems like maybe shrinking a font to half its intended size would be the minimum desired in most cases.  In the case of a single character, or emoji, shrinking until it fits (or the extreme 10% size case) may be desired.

iujames:
Ran the UI test suite on this, appears there will be one expected change in the `test-fixed-size` component test, which will now scale a bit to wrap, rather than ellipsize.

| After | Before |
| --- | ---- |
| failure_2_7928B3FB-CB92-4A86-BD19-C4F1CAE05640 | reference_1_7928B3FB-CB92-4A86-BD19-C4F1CAE05640 |

mmaatttt (approved):

Sources/AppcuesKit/Presentation/Extensions/View+Appcues.swift
iujames: @mmaatttt something like 
```
.minimumScaleFactor(min(10.0 / (model.style?.fontSize ?? 17.0), 1.0))
```
that would let things scale down to 10pt font, regardless of original size?

iujames: Thanks, was trying to find the right system prop to reference for that. updated!

PR #274, opened 2022-10-31
Fix border rendering

mmaatttt:
The general approach is now matching the CSS `border-box` model where `Content + Padding + Border = Width`.

There's a bunch of variable corner stuff _just_ for the case where there's differing padding values.


Sources/AppcuesKit/Presentation/Extensions/View+CornerRadius.swift
```diff
@@ -0,0 +1,77 @@
...
+//  AppcuesKit
+//
+//  Created by Matt on 2022-10-27.
+//  Copyright © 2022 Appcues. All rights reserved.
+//
+
+import SwiftUI
+
+@available(iOS 13.0, *)
+internal struct VariableRoundedCorner: Shape {
```
iujames, line 12: wow! nice job handling all the details here

iujames (approved):

Sources/AppcuesKit/Presentation/Traits/Appcues/DialogContainerView.swift
```diff
@@ -43,7 +44,7 @@ internal class DialogContainerView: UIView {
             dialogView.bottomAnchor.constraint(lessThanOrEqualTo: safeAreaLayoutGuide.bottomAnchor),
             dialogView.centerYAnchor.constraint(equalTo: centerYAnchor),
             // this is required so the dialogView has an initial non-zero height, after which it can start sizing to the content.
-            dialogView.heightAnchor.constraint(greaterThanOrEqualToConstant: 1),
+            dialogView.layoutMarginsGuide.bottomAnchor.constraint(greaterThanOrEqualTo: dialogView.layoutMarginsGuide.topAnchor, constant: 1),
```
mmaatttt, line 47: @iujames this line fixes the 0 height issue with carousels. What was happening is if the layout margins from the border took the 1px of space, then the `dialogView` would never get laid out. I've rewritten the constraint to ensure there's always the necessary 1px regardless of the layout margins.

iujames: 👍 Good find - tested that modal out locally here and it looks good now.

Sources/AppcuesKit/Presentation/Extensions/View+Appcues.swift
```diff
@@ -107,7 +107,7 @@ extension View {
                     .padding(width)
                     .overlay(
                     // Need to adjust the corner radius to match the radius applied to the view.
-                    RoundedRectangle(cornerRadius: (style.cornerRadius ?? 0) - width / 2)
+                    RoundedRectangle(cornerRadius: max(0, (style.cornerRadius ?? 0) - width / 2))
```
iujames, line 110: from my testing, it looked like we needed
```
cornerRadius + width / 2
```
actually, to match the radius of the view within that border?

given a border width 20, and a corner radius 10 on the view - this existing calculation gives:
Screen Shot 2022-11-01 at 1 45 06 PM
with some visible gaps in the corners

updating to the new calculation gives
Screen Shot 2022-11-01 at 1 45 42 PM

using a `max` here would also result in no radius applied to the corners any time the width is larger than 2x the radius, for example a corner radius of 5 and a border width of 20.

mmaatttt: This scenario is now handled properly:
https://github.com/appcues/appcues-mobile-experience-spec/blob/1fbbd2d22985f74133f0e5002e0a54f48687ff17/component-examples/image-corner-10-border-30.json
https://github.com/appcues/appcues-mobile-experience-spec/blob/1fbbd2d22985f74133f0e5002e0a54f48687ff17/verification-suites/ios/SnapshotTests/__Snapshots__/ImageComponentTests/testCorner10Border30.1.png

iujames (approved):
Looking good!

PR #275, opened 2022-11-02
Add support for the url dataType on textInput

mmaatttt:
As per the spec update:
- https://github.com/appcues/appcues-mobile-experience-spec/pull/110

I'm not sure specifically what difference `UITextContentType.URL` makes, but it seems like the right option.

Simulator Screen Shot - iPhone 14 Pro - 2022-11-02 at 09 46 50


iujames (approved):
Looks like it makes the `.` , `/` and `.com` more prominent, which seems appropriate.

mmaatttt:
@iujames that's the effect of `UIKeyboardType.URL`. `UITextContentType` generally influence autofill suggestions, but I'm not sure how that'd apply for a URL...

iujames:
ah sorry, my confusion, thanks.

PR #282, opened 2022-11-15
Provide more direct instruction for setting up the URL scheme during install

iujames:


mmaatttt (approved):
Much better! The "is not required" part was always a bit awkward.

PR #283, opened 2022-11-16
Add support for screen navigation prior to presenting new step container

iujames:
Draft of POC for feedback around supporting the "navigate" pre-step actions proposed here https://github.com/appcues/appcues-mobile-experience-spec/pull/118.

The changes in the SDK were fairly minimal. This also includes some updates in the SPM example to implement the AppcuesNavigationDelegate and related code restructuring to satisfy SwiftLint and little details there.

The primary logic flow is in the StateMachine `fromEndingStepToBeginningStep` - where it inspects the new step group just before starting the presentation of a new container (`presentContainer` side effect). If there are "navigate" actions, then they are passed along and executed, in sequence, prior to presenting the new container. This does _not_ occur on `fromBeginningExperienceToBeginningStep` - the first step - since the presentation on the first step should occur wherever the flow was targeted and qualified.

Some refactoring was done in `ActionRegistry` to support this new style of action execution, in addition to the couple already in there, but attempting to maintain as much shared logic flow with those code paths as possible. Part of this involved moving away from the queue processing in there and into a more direct recursive execution flow for sequential actions - so we can definitively know when an array of actions has fully completed and invoke a completion callback needed in this use case.

Also some updates in the `AppcuesLinkAction` to support the new concept of an optional `AppcuesNavigationDelegate` that can be used to execute links coming from experiences. This allows delegation of link handling to the host application, which can provide definitive completion handlers to indicate that navigation to a new screen is fully completed, prior to launching the next step in the experience.

I don't plan to merge this until we have more general consensus around how navigation steps will be handled in the builder and data model, but figured I'd post the work in progress in the meantime.

### DIFF
--- a/Examples/DeveloperSPMExample/SPMExample/DeepLinkNavigator.swift
+++ b/Examples/DeveloperSPMExample/SPMExample/DeepLinkNavigator.swift
@@ -53,7 +53,8 @@ class DeepLinkNavigator: AppcuesNavigationDelegate {
                 self.experienceID = components.experienceID
             } else if url.host == "appcues-mobile-links.netlify.app" {
                 // try to handle as a universal link to our associated domain
-                guard let destination = Destination(path: components.path) else { return nil }
+                let path = components.path.trimmingCharacters(in: CharacterSet(charactersIn: "/"))
+                guard let destination = Destination(path: path) else { return nil }
                 self.destination = destination
                 self.experienceID = components.experienceID
             } else {
--- a/Examples/DeveloperSPMExample/SPMExample/DeepLinkNavigator.swift
+++ b/Examples/DeveloperSPMExample/SPMExample/DeepLinkNavigator.swift
@@ -104,13 +104,23 @@ class DeepLinkNavigator: AppcuesNavigationDelegate {
     }
 
     private func handle(url: URL?, completion: ((Bool) -> Void)?) {
-        guard
-            let url = url,
-            let target = DeepLink(url: url),
-            let windowScene = scene as? UIWindowScene,
+        guard let url = url else {
+            // no valid URL given, cannot navigate
+            completion?(false)
+            return
+        }
+
+        guard let target = DeepLink(url: url) else {
+            // the link was not a known deep link for this application, so pass along off to OS to handle
+            UIApplication.shared.open(url, options: [:]) { success in completion?(success) }
+            return
+        }
+
+        guard let windowScene = scene as? UIWindowScene,
             let window = windowScene.windows.first(where: { $0.isKeyWindow }),
             let origin = AppScreen(rootController: window.rootViewController)
         else {
+            // cannot find the screen information to navigate, fail navigation
             completion?(false)
             return
         }
--- a/Examples/DeveloperSPMExample/SPMExample/DeepLinkNavigator.swift
+++ b/Examples/DeveloperSPMExample/SPMExample/DeepLinkNavigator.swift
@@ -10,33 +10,42 @@ import AppcuesKit
 
 // NOTE: This deep link implementation should not be taken as an example of best practices. It's not particularly scalable,
 // it's coupled to the specific view controller hierarchy of the app, and uses segues which can cause runtime crashes.
-class DeepLinkNavigator {
+class DeepLinkNavigator: AppcuesNavigationDelegate {
 
     /// Link destinations
-    enum DeepLink {
-        case signIn
-        case events
-        case profile
-        case group
+    struct DeepLink {
+        enum Destination {
+            case signIn
+            case events
+            case profile
+            case group
 
-        var tabIndex: Int? {
-            // needs to match Main.storyboard
-            switch self {
-            case .signIn: return nil
-            case .events: return 0
-            case .profile: return 1
-            case .group: return 2
+            var tabIndex: Int? {
+                // needs to match Main.storyboard
+                switch self {
+                case .signIn: return nil
+                case .events: return 0
+                case .profile: return 1
+                case .group: return 2
+                }
             }
         }
 
-        init?(host: String?) {
-            switch host?.lowercased() {
-            case "signin": self = .signIn
-            case "events": self = .events
-            case "profile": self = .profile
-            case "group": self = .group
+        let destination: Destination
+        let experienceID: String?
+
+        init?(url: URL) {
+            guard url.scheme == "appcues-example" else { return nil }
+
+            switch url.host?.lowercased() {
+            case "signin": destination = .signIn
+            case "events": destination = .events
+            case "profile": destination = .profile
+            case "group": destination = .group
             default: return nil
             }
+
+            experienceID = url.experienceID
         }
     }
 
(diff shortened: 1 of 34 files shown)


Examples/DeveloperSPMExample/SPMExample/DeepLinkNavigator.swift
```diff
@@ -10,33 +10,42 @@ import AppcuesKit
 
 // NOTE: This deep link implementation should not be taken as an example of best practices. It's not particularly scalable,
 // it's coupled to the specific view controller hierarchy of the app, and uses segues which can cause runtime crashes.
-class DeepLinkNavigator {
+class DeepLinkNavigator: AppcuesNavigationDelegate {
```
mmaatttt, line 13: FYI I realized I changed the `DeeplinkNavigator` in the CocoapodsExample example for universal links, but not the SPM one. It'd be good to eventually reconcile them. I'm happy to do that when the time comes.

Sources/AppcuesKit/Presentation/Actions/Appcues/AppcuesLinkAction.swift
```diff
@@ -46,12 +46,25 @@ internal class AppcuesLinkAction: ExperienceAction {
                 completion()
             } else {
                 if openExternally {
-                    urlOpener.open(url, options: [:]) { _ in completion() }
+                    openLink(appcues: appcues, completion: completion)
```
mmaatttt, line 49: When I was just thinking about the concept, I wasn't sure how we'd get access to the `navigationDelegate` in the link action. Turns out that's easy 👌

Sources/AppcuesKit/Presentation/ExperienceRendering/ExperienceStateMachine.swift
```diff
@@ -222,8 +222,13 @@ extension ExperienceStateMachine {
...
                 try machine.transition(action)
-            case let .presentContainer(experience, stepIndex, package):
-                executePresentContainer(machine: machine, experience: experience, stepIndex: stepIndex, package: package)
+            case let .presentContainer(experience, stepIndex, package, actions):
+                machine.actionRegistry.enqueue(actionModels: actions) {
+                    executePresentContainer(machine: machine,
+                                            experience: experience,
+                                            stepIndex: stepIndex,
+                                            package: package)
+                }
```
mmaatttt, line 231: So simple and so good!

Sources/AppcuesKit/Presentation/ExperienceRendering/ExperienceStateMachine+State.swift
```diff
@@ -193,11 +193,15 @@ extension ExperienceStateMachine.Transition {
                 sideEffect: .error(.step(experience, currentIndex, "Step at \(stepRef) does not exist"), reset: false))
         }
 
+        let stepGroup = experience.steps[stepIndex.group]
+        let navigationActions = stepGroup.actions[stepGroup.id.uuidString.lowercased()]?.filter { $0.trigger == "navigate" } ?? []
```
mmaatttt, line 197: Unrelated to the main idea here, but this reminds me: we've done the `.lowercased()` thing enough times that we should have an extension on the UUID type that gets a lowercased "appcues compatible" version of a UUID string.

Sources/AppcuesKit/Presentation/ExperienceRendering/ExperienceStateMachine.swift
```diff
@@ -222,8 +222,13 @@ extension ExperienceStateMachine {
             switch self {
             case .continuation(let action):
                 try machine.transition(action)
-            case let .presentContainer(experience, stepIndex, package):
-                executePresentContainer(machine: machine, experience: experience, stepIndex: stepIndex, package: package)
+            case let .presentContainer(experience, stepIndex, package, actions):
+                machine.actionRegistry.enqueue(actionModels: actions) {
```
iujames, line 226: this is where the pre-step actions execute before the container presentation on line 227, which happens after action completion.

Sources/AppcuesKit/Presentation/Actions/ActionRegistry.swift
```diff
@@ -99,4 +87,18 @@ internal class ActionRegistry {
...
+        guard let appcues = appcues, !models.isEmpty else {
+            completion?()
+            return
+        }
+
+        let next = models.removeFirst()
+        next.execute(inContext: appcues) {
+            self.execute(models, completion: completion)
+        }
+    }
```
mmaatttt, line 103: Not sure if it matters, but I think there is a subtle change here from how it was? Previously actions from, say, 3 different buttons would all be enqueued sequentially and executed one at a time. I think now we'd get 3 parallel `execute` calls going at once? Like I said, not sure if it matters, or even which is preferable.

I think if we wanted to keep the old behaviour, we could maybe add the completion block as a pseudo-action in the queue after the actions it enqueued so it would get called in place:

```
- BUTTON1.ACTION1
- BUTTON1.ACTION2
- BUTTON1.COMPLETION
- BUTTON2.ACTION1
- BUTTON2.COMPLETION
```

mmaatttt (commented):
This is great! Will hold the approval until we're confirmed on this approach, but it's all good from my POV

Sources/AppcuesKit/Presentation/Actions/ActionRegistry.swift
iujames: Ah good catch. Yeah I think that sort of approach would work fine to keep the previous behavior. I can look into that before moving this forward.

mmaatttt: The downside of the previous approach is one bad action (ie one that doesn't call its completion handler) can forever gum up the queue 🤷‍♂️

iujames: There is a similar implementation on Android, using an Action queue built on a Kotlin Channel. We'll need to invent some way to get a completion notification back from there as well. Going to see if a similar idea can work for both.

iujames: @mmaatttt went ahead and reverted back the previous queue handling for the interactive actions case, so everything functions the same as before for actions triggered by user tap, etc. The only new thing here now is the `enqueue` with completion, which calls `execute` recursively. This one is only used for the new pre-step container present actions.

This matches how things were [handled on Android](https://github.com/appcues/appcues-android-sdk/pull/271) as well, and in general, I like the concept of all interactive actions flowing through a global sequential queue.

As found on Android, it doesn't work well to include the new "navigate" actions into the queue and try to await completion, since the execution of those actions actually occurs within a currently executing Action (the continue to next step typically) - and ends up deadlocking the step transition, as that continue action is never able to fully complete if something inside of it is waiting.

Sources/AppcuesKit/Presentation/Actions/Appcues/AppcuesLinkAction.swift
iujames: @mmaatttt  - maybe this line 49 should continue to just launch the link through the UIApplication (the old urlOpener approach)? And line 55 would be the only line that calls into the `navigationDelegate`. I say this because I realize that at this line, we already know it is an http link (not a scheme link), and it is not a universal link (per line 43), and it's being requested to openExternally. This is just adding extra burden on the host application to conditionally check for web links and open into the browser when we could do it here.

mmaatttt: True, true. Good catch!

mmaatttt (approved):
🎉

PR #285, opened 2022-11-28
Add configuration option to control universal link handling

iujames:
This approach would allow a customer to opt-out of the universal link handling logic, if their app was not compatible with this approach. The issue here arose in Ionic apps, and we can update our Capacitor plugin to disable this feature, by default, to avoid common issues.


iujames:
Will add a test for this before merge.

PR #287, opened 2022-11-30
Rework experiment analytics to support new requested properties

iujames:
* capture 2 new properties from the response for experiments
* update the `experience_entered` analytic to output 3 additional properties and rename another

Also added extension on `UUID` for string formatting, as originally mentioned in https://github.com/appcues/appcues-ios-sdk/pull/283#discussion_r1024438630

### DIFF
--- a/Sources/AppcuesKit/Data/Models/Experiment.swift
+++ b/Sources/AppcuesKit/Data/Models/Experiment.swift
@@ -12,6 +12,8 @@ internal struct Experiment {
     let group: String
     let experimentID: UUID
     let experienceID: UUID
+    let goalID: String
+    let contentType: String
 
     var shouldExecute: Bool {
         return group != "control"
@@ -23,5 +25,7 @@ extension Experiment: Decodable {
         case group
         case experimentID = "experimentId"
         case experienceID = "experienceId"
+        case goalID = "goalId"
+        case contentType
     }
 }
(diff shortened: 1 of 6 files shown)


Sources/AppcuesKit/Data/Models/Experiment.swift
```diff
@@ -12,6 +12,8 @@ internal struct Experiment {
     let group: String
     let experimentID: UUID
     let experienceID: UUID
+    let goalID: String
```
iujames, line 15: per Nate, this needs to be a string, cannot be assumed a UUID due to legacy goal usage in Firebase

Sources/AppcuesKit/Data/Models/Experiment.swift
```diff
@@ -23,5 +25,7 @@ extension Experiment: Decodable {
         case group
         case experimentID = "experimentId"
         case experienceID = "experienceId"
+        case goalID = "goalId"
+        case contentType
```
iujames, line 29: the JSON keys are `goal_id` and `content_type` in the response, but appear as camelCase to us, due to https://github.com/appcues/appcues-ios-sdk/blob/main/Sources/AppcuesKit/Data/Networking/NetworkClient.swift#L124

Sources/AppcuesKit/Presentation/ExperienceRendering/LifecycleEvent.swift
```diff
@@ -46,15 +46,15 @@ internal enum LifecycleEvent: String, CaseIterable {
...
+            properties["stepId"] = step.id.appcuesFormatted
             properties["stepType"] = step.type
             // stepIndex is added primarily for use by the debugger
             properties["stepIndex"] = stepIndex.description
         }
 
         if let error = error {
             properties["message"] = error.message
-            properties["errorId"] = error.id.uuidString
+            properties["errorId"] = error.id.appcuesFormatted
```
iujames, line 57: This one was not `lowercased()` before, but I'm assuming it should be.

mmaatttt: This one doesn't matter since its generated on the client, but it doesn't hurt to consistent with what gets set to the backend.

mmaatttt (approved):

PR #288, opened 2022-12-02
Support lossy decoding of Experience objects in qualify response

iujames:
The general concept is very similar to what was done on Android https://github.com/appcues/appcues-android-sdk/pull/278
* `QualifyResponse` now has `[LossyExperience]` instead of `[Experience]`
* There are two possible types - a normal `.decoded(Experience)` and a `.failed(FailedExperience)`
* `FailedExperience` captures minimal info that could hopefully be used to report flow issues for troubleshooting later
* custom deserialization in the `QualifyResponse` does the attempted `Experience` deserialization on each item, error handling and capturing detailed error message, and then fallback handling for failures

an example failure message might look like this for the case of a missing required key
```
key 'CodingKeys(stringValue: "type", intValue: nil)' not found: No value associated with key CodingKeys(stringValue: "type", intValue: nil) ("type"). codingPath: [_JSONKey(stringValue: "Index 0", intValue: 0), CodingKeys(stringValue: "steps", intValue: nil), _JSONKey(stringValue: "Index 0", intValue: 0), CodingKeys(stringValue: "traits", intValue: nil), _JSONKey(stringValue: "Index 2", intValue: 2)]
```
The generated output from the decoding errors is a bit gnarly, but this can be interpreted as: the key `type` was not found. The path it was expected was: index zero of the `experiences` array, in the `steps` array at index zero, in the `traits` collection at index 2.

### DIFF
--- a/Sources/AppcuesKit/Data/Analytics/AnalyticsTracker.swift
+++ b/Sources/AppcuesKit/Data/Analytics/AnalyticsTracker.swift
@@ -117,7 +117,7 @@ internal class AnalyticsTracker: AnalyticsTracking, AnalyticsSubscribing {
     private func process(qualifyResponse: QualifyResponse, activity: Activity) {
         if let experienceRenderer = container?.resolve(ExperienceRendering.self) {
             let experiments = qualifyResponse.experiments ?? []
-            let qualifiedExperienceData: [ExperienceData] = qualifyResponse.experiences.map { experience in
+            let qualifiedExperienceData: [ExperienceData] = qualifyResponse.experiences.parsed().map { experience in
                 let experiment = experiments.first { $0.experienceID == experience.id }
                 return ExperienceData(experience,
                                       priority: qualifyResponse.renderPriority,
--- a/Sources/AppcuesKit/Data/Extensions/UUID+Custom.swift
+++ b/Sources/AppcuesKit/Data/Extensions/UUID+Custom.swift
@@ -12,11 +12,11 @@ import Foundation
 extension UUID {
     static var generator: () -> UUID = UUID.init
 
-    static func create() -> UUID {
-        return UUID.generator()
-    }
-
     var appcuesFormatted: String {
         return uuidString.lowercased()
+    }
+
+    static func create() -> UUID {
+        return UUID.generator()
     }
 }
--- a/Sources/AppcuesKit/Data/Models/Experience.swift
+++ b/Sources/AppcuesKit/Data/Models/Experience.swift
@@ -170,3 +170,16 @@ extension Experience.Action: Decodable {
     }
 
 }
+
+// utility to quickly extract the list of valid Experiences from an array of
+// LossyExperience values
+extension Array where Element == LossyExperience {
+    func parsed() -> [Experience] {
+        return self.compactMap {
+            if case let .decoded(experience) = $0 {
+                return experience
+            }
+            return nil
+        }
+    }
+}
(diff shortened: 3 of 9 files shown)


Sources/AppcuesKit/Data/Extensions/UUID+Custom.swift
```diff
@@ -12,11 +12,11 @@ import Foundation
...
-        return UUID.generator()
-    }
-
     var appcuesFormatted: String {
         return uuidString.lowercased()
     }
+
+    static func create() -> UUID {
+        return UUID.generator()
+    }
```
iujames, line 21: moving this to fix SwiftLint issue from prior change

Sources/AppcuesKit/Data/Analytics/AnalyticsTracker.swift
```diff
@@ -117,7 +117,7 @@ internal class AnalyticsTracker: AnalyticsTracking, AnalyticsSubscribing {
     private func process(qualifyResponse: QualifyResponse, activity: Activity) {
         if let experienceRenderer = container?.resolve(ExperienceRendering.self) {
             let experiments = qualifyResponse.experiments ?? []
-            let qualifiedExperienceData: [ExperienceData] = qualifyResponse.experiences.map { experience in
+            let qualifiedExperienceData: [ExperienceData] = qualifyResponse.experiences.parsed().map { experience in
```
iujames, line 120: Here is where we only take the properly parsed items currently, not the failures. In the future we can try to do something with the failed experiences to log flow issues.

Sources/AppcuesKit/Data/Models/QualifyResponse.swift
```diff
@@ -35,10 +35,44 @@ extension QualifyResponse: Decodable {
...
+                }
+
+                // if we get here, it means we failed normal decoding. Try to decode the minimal
+                // info needed to report a flow issue that we can troubleshoot
+                if var element = try? experienceContainer.decode(FailedExperience.self) {
+                    element.error = errorMessage
+                    decodedExperiences.append(.failed(element))
+                } else {
+                    // cannot decode anything at all about this experience, skip it
+                    try experienceContainer.skip()
```
mmaatttt, line 72: I thought this was kinda gross at first, but reading up a bit and it seems like the only way.

Sources/AppcuesKit/Data/Models/Experience.swift
```diff
@@ -15,6 +15,19 @@ internal protocol StepModel {
...
 
+internal enum LossyExperience {
+    case decoded(Experience)
+    case failed(FailedExperience)
+}
+
+internal struct FailedExperience: Decodable {
+    let id: UUID
+    let name: String?
+    let type: String?
```
mmaatttt, line 26: Thinking ahead a bit, the easiest way to log errors is in the ExperienceRenderer (specifically `.startExperience` in the state machine). That requires an `ExperienceData` (and I don't want to change that) which in turn has an `Experience` object (and changing this in `ExperienceData` to optional or something would be a headache).

If we make `FailedExperience` here require `name: String` and `type: String` rather than have those optional, we can manually construct a skeleton Experience object out of that that would fit in `ExperienceData`. Then adding an option `error: String?` to `ExperienceData` would let the state machine check for that in `.startExperience` and then report it.

This should be pretty easy, with the only change being that we're requiring two more things in `FailedExperience` (which should be there every time and would be skipped otherwise). All the decoding errors I'm worried about (read: interested in) are within the content of the experience, not the highest level structure.

mmaatttt (approved):
Good stuff!

Sources/AppcuesKit/Data/Models/Experience.swift
iujames: yep, I can take a stab at it (ticket 45190)

PR #294, opened 2022-12-08
Allow navigate actions on first step for non-qualifying experience triggers

iujames:
Stacks on other step navigation work in #283 

Analogous to what was done on Android side in https://github.com/appcues/appcues-android-sdk/pull/281

This PR introduces the concept of ExperienceTrigger, which defines the currently six ways an Experience can be triggered:

* builder preview
* qualification
* deep link
* show function call
* launch experience action
* post-flow completion action

This context is now stored on the `ExperienceData` object via passing through at call sites as appropriate. This allows us to detect when an Experience is being launched from a method other than the standard qualification flow, and enable pre-step navigation actions to occur if so. Qualified flows are unchanged - as the location in the app they trigger upon is based entirely on the flow settings defined in Studio, and no redirection is done on launch in those cases.

This will enable use cases like redirecting to the right page prior to showing the experience, when using the builder preview, or a deep link to a flow.

Here is an example where the app is on the Events tab, and a deep link to preview a flow with the first step targeting the Profile tab results in the app switching to the correct tab and showing the flow.

https://user-images.githubusercontent.com/19266448/206551578-8f8794de-5e00-4aeb-a420-46f84ddb0349.mov

### DIFF
--- a/Sources/AppcuesKit/Presentation/ExperienceRendering/ExperienceStateMachine+State.swift
+++ b/Sources/AppcuesKit/Presentation/ExperienceRendering/ExperienceStateMachine+State.swift
@@ -153,16 +153,12 @@ extension ExperienceStateMachine.Transition {
         do {
             let package = try traitComposer.package(experience: experience, stepIndex: stepIndex)
 
-            // for pre-step navigation actions - only allow these to execute if this experience is being launched for some
-            // other reason than qualification (i.e. deep links, preview, manual show). For any qualified experience, the initial
-            // starting state of the experience is determined solely by flow settings determining the trigger
-            // (i.e. trigger on certain screen).
             let navigationActions: [Experience.Action]
-            if case .qualification = experience.trigger {
-                navigationActions = []
-            } else {
+            if experience.trigger.shouldNavigateBeforeRender {
                 let stepGroup = experience.steps[stepIndex.group]
                 navigationActions = stepGroup.actions[stepGroup.id.appcuesFormatted]?.filter { $0.trigger == "navigate" } ?? []
+            } else {
+                navigationActions = []
             }
 
             return .init(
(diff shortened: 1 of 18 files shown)


Sources/AppcuesKit/Presentation/ExperienceRendering/ExperienceStateMachine+State.swift
```diff
@@ -152,9 +152,22 @@ extension ExperienceStateMachine.Transition {
         let stepIndex = Experience.StepIndex.initial
         do {
             let package = try traitComposer.package(experience: experience, stepIndex: stepIndex)
+
+            // for pre-step navigation actions - only allow these to execute if this experience is being launched for some
+            // other reason than qualification (i.e. deep links, preview, manual show). For any qualified experience, the initial
+            // starting state of the experience is determined solely by flow settings determining the trigger
+            // (i.e. trigger on certain screen).
+            let navigationActions: [Experience.Action]
+            if case .qualification = experience.trigger {
```
mmaatttt, line 161: I don't usually like using non-exhaustive conditionals for enums since it can be  hard to know what needs to be updated if we added another `ExperienceTrigger` case, especially when it's not a property of the enum itself. So for this, I think a computed property on `ExperienceTrigger` like `shouldProcessNavigation` would work nicely.

mmaatttt (approved):

Sources/AppcuesKit/Presentation/ExperienceRendering/ExperienceTrigger.swift
```diff
@@ -0,0 +1,18 @@
...
+//  AppcuesKit
+//
+//  Created by James Ellis on 12/8/22.
+//  Copyright © 2022 Appcues. All rights reserved.
+//
+
+import Foundation
+
+internal enum ExperienceTrigger {
+    case qualification(reason: String?)
```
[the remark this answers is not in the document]
iujames: No reason to not use the enum - nice improvement!

This code actually dates back to some previous discussions around tracking analytics for how flows are triggered for user journey mapping. This associated value is not used at all currently - so I'm not opposed to just taking it out under the rules of YAGNI.

However, it seemed fairly likely that similar future needs for this type of data would come up, so that's why I kept it along for the ride.

Sources/AppcuesKit/Presentation/ExperienceRendering/ExperienceStateMachine+State.swift
iujames: Good call - agree that is a better approach, updated.

PR #295, opened 2022-12-09
Enforce single instance of each trait type at each level

iujames:
targeting `traits-2.0`

This is part of the 2.0 work where we've made the decision to enforce that a certain trait type, i.e. `@appcues/backdrop`, can only have a single instance at any given level of the experience (typically group or step level definitions).

* Added custom decoding in `TraitCollection` object that detects duplicate trait `type` keys at any given level, and throws a `DecodingError` if encountered.
* Updated to use in `Experience`, `Experience.Step.Group` and `Experience.Step.Child` decoding
* Tests to validate each error case and normal non-error case

If the error is detected, it will result in an Experience error like shown here
Screenshot 2022-12-09 at 10 38 41 AM

### DIFF
--- a/Sources/AppcuesKit/Data/Models/Experience.swift
+++ b/Sources/AppcuesKit/Data/Models/Experience.swift
@@ -51,6 +51,44 @@ internal struct FailedExperience: Decodable {
     }
 }
 
+// a helper that deserializes a collection of traits and also enforces that
+// no trait types are duplicated within the collection
+internal struct TraitCollection: Decodable {
+    var traits: [Experience.Trait]
+
+    init(from decoder: Decoder) throws {
+        var container = try decoder.unkeyedContainer()
+
+        var traits: [Experience.Trait] = []
+        var traitTypes = Set<String>()
+        var duplicates = Set<String>()
+
+        if let count = container.count {
+            traits.reserveCapacity(count)
+        }
+
+        while !container.isAtEnd {
+            let trait = try container.decode(Experience.Trait.self)
+            if traitTypes.insert(trait.type).inserted {
+                // normal case, capture the decoded trait for our resulting array
+                traits.append(trait)
+            } else {
+                // a dupe was found and this will end up causing a decoding error below
+                duplicates.insert(trait.type)
+            }
+        }
+
+        if !duplicates.isEmpty {
+            let message = "multiple traits of same type are not supported: \(duplicates.joined(separator: ","))"
+            throw DecodingError.dataCorrupted(
+                DecodingError.Context(codingPath: decoder.codingPath,
+                                      debugDescription: message))
+        }
+
+        self.traits = traits
+    }
+}
+
 internal struct Experience {
 
     @dynamicMemberLookup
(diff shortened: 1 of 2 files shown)


Sources/AppcuesKit/Data/Models/Experience.swift
```diff
@@ -51,6 +51,44 @@ internal struct FailedExperience: Decodable {
...
+        }
+
+        while !container.isAtEnd {
+            let trait = try container.decode(Experience.Trait.self)
+            if traitTypes.insert(trait.type).inserted {
+                // normal case, capture the decoded trait for our resulting array
+                traits.append(trait)
+            } else {
+                // a dupe was found and this will end up causing a decoding error below
+                duplicates.insert(trait.type)
```
mmaatttt, line 77: I like that we're collecting all duplicates rather than failing on the first one 👍

Sources/AppcuesKit/Data/Models/Experience.swift
```diff
@@ -115,6 +153,18 @@ extension Experience: Decodable {
         case redirectURL = "redirectUrl"
         case nextContentID = "nextContentId"
     }
+
+    init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        id = try container.decode(UUID.self, forKey: .id)
+        name = try container.decode(String.self, forKey: .name)
+        type = try container.decode(String.self, forKey: .type)
+        publishedAt = try? container.decode(Int.self, forKey: .publishedAt)
```
mmaatttt, line 162: Interesting thing I discovered recently: In Xcode 14, you can cmd+click on a Codable type and choose `Add Explicit Codable Implementation`. For something like `let optionalValue: String?`, it generates `try container.decodeIfPresent` rather than `try? container.decode`. I wasn't familiar with `decodeIfPresent`, but the [docs](https://developer.apple.com/documentation/swift/keyeddecodingcontainer/decodeifpresent(_:forkey:)-2hn6i) are insightful:

> A decoded value of the requested type, or nil if the Decoder does not have an entry associated with the given key, or if the value is a null value.

So the nuance being that JSON where the key exists with the wrong value type (eg `"optionalValue": true`) would fail the decoding with an error for `try container.decodeIfPresent` but just set a nil value with `try? container.decode` because `try?` swallows the error. Essentially, `decodeIfPresent` is a stronger check than `try? container.decode`.

The default `Decodable` conformance before this change uses `decodeIfPresent`, so technically this is weakening the model decoding validation. I wouldn't say that's a problem (or that this needs to be changed), but it is something to be aware of with all the usages of `try?` in this PR.

As another example I just looked at the places where we currently use `try?` in decoders, and there's `qualificationReason = try? container.decode(QualificationReason.self, forKey: .qualificationReason)` where `try?` would be the right usage because don't want an error for unknown qualificationReason values.

Tests/AppcuesKitTests/Networking/ExperienceDecodeTests.swift
```diff
@@ -0,0 +1,259 @@
...
+        do {
+            experience = try NetworkClient.decoder.decode(Experience.self, from: data)
+        } catch let DecodingError.dataCorrupted(context) {
+            error = "\(context)"
+        }
+
+        // Assert
+        XCTAssertNil(experience)
+        XCTAssertNotNil(error)
+        XCTAssertTrue(error!.contains("multiple traits of same type are not supported: step-trait"))
```
mmaatttt, line 135:
```swift
let unwrappedError = try XCTUnwrap(error)
XCTAssertTrue(unwrappedError.contains("multiple traits of same type are not supported: step-trait"))

// or even
XCTAssertTrue(try XCTUnwrap(error).contains("multiple traits of same type are not supported: step-trait"))
```
is a slightly better option since it saves the test from crashing on line 135 if error is actually `nil`. Doesn't matter too much, but it is nicer if the test ever does fail.

mmaatttt (approved):

Sources/AppcuesKit/Data/Models/Experience.swift
iujames: That is good to understand the difference here in the options - thanks for the info! I scanned through `try?` usages here again and didn't really see any I'd want stricter behavior on either. But let me know if anything else jumps out that seems worth of improving along those lines.

Tests/AppcuesKitTests/Networking/ExperienceDecodeTests.swift
iujames: nice, thanks for the tip on `XCTUnwrap` - updated these

Sources/AppcuesKit/Data/Models/Experience.swift
mmaatttt: Agreed that `try?` is appropriate for the uses in this PR 👍

PR #296, opened 2022-12-12
Fix Debugger expanded view layout on iPad

iujames:
Side note, noticed that `NavigationView` is on the way out, in favor of `NavigationStack` and `NavigationSplitView`, but that is iOS 16+ minimum. https://developer.apple.com/documentation/swiftui/migrating-to-new-navigation-types

| after | before |
| ---- | ---- |
| Simulator Screen Shot - iPad Pro (12 9-inch) (6th generation) - 2022-12-12 at 10 13 26  | Simulator Screen Shot - iPad Pro (12 9-inch) (6th generation) - 2022-12-12 at 10 06 54 |

### DIFF
--- a/Sources/AppcuesKit/Presentation/Debugger/Panel/DebugUI.swift
+++ b/Sources/AppcuesKit/Presentation/Debugger/Panel/DebugUI.swift
@@ -49,6 +49,7 @@ internal enum DebugUI {
                 .navigationBarTitle(Text("Debugger"))
                 .navigationBarHidden(true)
             }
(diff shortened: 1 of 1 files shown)


mmaatttt (approved):
>Side note, noticed that NavigationView is on the way out, in favor of NavigationStack and NavigationSplitView, but that is iOS 16+ minimum

Gotta love SwiftUI..."It's declarative, so you don't need to worry about the implementation details...EXCEPT" 🙃 `NavigationStack` does look though.

PR #297, opened 2022-12-13
Update how config values from the JSON model are provided in Trait and Action init

mmaatttt:
There's an irritating annoyance with our current trait inits: Currently it's `init?(config: [String: Any]?, level: ExperienceTraitLevel)` where the config dictionary decodes some types (String, Bool, Int, Double) and the rest are left in a coding container where we can do the clever like `config?["content", decodedAs: ExperienceComponent.self]`.

The problem is for our new `@appcues/backdrop-keyhole` we have a `cornerRadius` property which should be a Double (since 2.5 is a perfectly acceptable value), but often will have whole number values (eg 10). The decoding will make this different types depending on if there’s a decimal which is quite annoying because the trait init needs to handle both options despite the fact it knows that it should always decode as a Double. And from the perspective of a custom trait implementation, this is entirely non-standard behaviour and would be quite confusing to understand.

This PR changes it where we don't automatically decode anything and leave it all up to the trait where the type info is fully known.

Also we don't actually do the partial dictionary decode for actions—only traits—so this PR makes both consistent.

I'm not entirely happy with the `DecodingExperienceConfig` name, so definitely open to ideas for that.

### Benefits:
- This is more performant since we don't try decoding every key to a bunch of types that it’ll never be.
- It's also nicer where a trait wants to decode to a Decodable enum (like the new `@appcues/step-transition-animation` trait), but the value has already been decoded to a string (I have the ugly `self.easing = Easing(rawValue: config?["easing"] as? String ?? "")` when it could be `self.easing = config["easing"]` with some clever type inference).
- A custom trait doesn't need to know that a `String` should be accessed differently from a `CustomType` because it's all the same now.

This is a :boom: breaking change for traits 2.0, but the end result would be much cleaner in the trait implementations and would require less explanation in the documentation about how “some values are already decoded for you”.

## Testing

The tests are actually the trickiest part of this change. They rely extensively on the `config: [String: Any]` init to configure test cases and even pass `XCTestExpectation` objects into custom traits. I wanted to keep the tests as simple as possible, and I managed to do that by created a way to convert a `[String: Any]` to the `[String: KeyedDecodingContainer<JSONCodingKeys>]` dictionary that the `DecodingExperienceConfig` now requires. There's a few complex pieces here:

1. An extension on `DecodingExperienceConfig` that takes a `[String: Any]` and converts it using `.toDecodableDict()`
2. `toDecodableDict()` creates a `FakeDecodingContainer: KeyedDecodingContainerProtocol` that just stores the value and returns it when "decoding" (rather than trying to encode and decode, since there's no guarantee all types are Encodable)
3. All the types do need to be `Decodable` and `XCTestExpectatation` isn't, so I created the new `DecodableExpectation` type that wraps an expectation and encodes just a simple `UUID` that then can be decoded and looked up in a static dictionary to get the original expectation back. Don't worry, I'd only ever consider this sort of wild hack in a test case 😅 

Consequently, most of the test changes are just wrapping the existing `[String: Any]` dictionaries in `DecodingExperienceConfig(...)`. And anything that's testing with `Experience.Action` uses `toDecodableDict()` directly.

### DIFF
--- a/Sources/AppcuesKit/Data/Networking/DecodingExperienceConfig.swift
+++ b/Sources/AppcuesKit/Data/Networking/DecodingExperienceConfig.swift
@@ -1,0 +1,51 @@
+//
+//  DecodingExperienceConfig.swift
+//  AppcuesKit
+//
+//  Created by Matt on 2022-12-12.
+//  Copyright © 2022 Appcues. All rights reserved.
+//
+
+import Foundation
+
+/// Object that stores the configuration options provided in ``Experience.Trait`` and ``Experience.Action`` models.
+public class DecodingExperienceConfig: NSObject {
+    // this is a class only for @objc compatibility
+
+    private let properties: [String: KeyedDecodingContainer<JSONCodingKeys>]
+
+    init(_ properties: [String: KeyedDecodingContainer<JSONCodingKeys>]?) {
+        self.properties = properties ?? [:]
+    }
+
+    /// Accesses the value associated with the given key for reading.
+    ///
+    /// The value will be decoded from the configuration to the expected type.
+    public subscript<T: Decodable>(_ key: String) -> T? {
+        guard let container = properties[key], let key = JSONCodingKeys(stringValue: key) else { return nil }
+
+        return try? container.decode(T.self, forKey: key)
+    }
+}
+
+extension DecodingExperienceConfig {
+    // Map the properties for the @appcues/update-profile action
+    var safeValues: [String: Any] {
+        properties.reduce(into: [:]) { dict, pair in
+            guard let key = JSONCodingKeys(stringValue: pair.key) else { return }
+            let container = pair.value
+
+            if let boolValue = try? container.decode(Bool.self, forKey: key) {
+                dict[key.stringValue] = boolValue
+            } else if let stringValue = try? container.decode(String.self, forKey: key) {
+                dict[key.stringValue] = stringValue
+            } else if let intValue = try? container.decode(Int.self, forKey: key) {
+                dict[key.stringValue] = intValue
+            } else if let doubleValue = try? container.decode(Double.self, forKey: key) {
+                dict[key.stringValue] = doubleValue
+            } else {
+                // not a supported type
+            }
+        }
+    }
+}
--- a/Tests/AppcuesKitTests/DecodingExperienceConfig+Testable.swift
+++ b/Tests/AppcuesKitTests/DecodingExperienceConfig+Testable.swift
@@ -1,0 +1,174 @@
+//
+//  DecodingExperienceConfig+Testable.swift
+//  AppcuesKitTests
+//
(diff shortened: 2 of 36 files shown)


Sources/AppcuesKit/Data/Networking/DecodingExperienceConfig.swift
```diff
@@ -0,0 +1,51 @@
...
+//  AppcuesKit
+//
+//  Created by Matt on 2022-12-12.
+//  Copyright © 2022 Appcues. All rights reserved.
+//
+
+import Foundation
+
+/// Object that stores the configuration options provided in ``Experience.Trait`` and ``Experience.Action`` models.
+public class DecodingExperienceConfig: NSObject {
```
iujames, line 12: checking understanding - is this basically just to provide a convenience wrapper type with helpers/extensions rather than having everything work with `[String: KeyedDecodingContainer<JSONCodingKeys>]` directly? It could be done just using type type or a `typedef` perhaps? Granted, this probably simplifies things.

iujames: re: naming, given some other chat, maybe `ExperiencePluginConfig`? `AppcuesPluginConfig`. Nothing really jumps out to me.

Tests/AppcuesKitTests/DecodingExperienceConfig+Testable.swift
```diff
@@ -0,0 +1,174 @@
...
+//  Created by Matt on 2022-12-12.
+//  Copyright © 2022 Appcues. All rights reserved.
+//
+
+import Foundation
+import XCTest
+@testable import AppcuesKit
+
+// Allows initializing a DecodingExperienceConfig from a test with a plain dictionary
+extension DecodingExperienceConfig {
```
iujames, line 14: Just wondering if there is any way to put `DecodingExperienceConfig` behind some Protocol that would allow for more simplified mocking in test, not needing to actually create any `KeyedDecodingContainer` etc. The subscript relies on T being Decodable though, so maybe not.

iujames (approved):
Looks good to me to simplify/standardize the typical access to config elements from traits and actions.

Chatted in Slack, other idea was to perhaps use something where traits/actions define their own config types (Decodables) which can then be used to fetch a fully formed config instance for these "plugin" types from the underlying config JSON - rather then picking out individual items by string key.

Good with progress here though and maybe we'll cycle back to other ideas when time permits.

Tests/AppcuesKitTests/DecodingExperienceConfig+Testable.swift
mmaatttt: Yeah exactly—the generic subscript complicates things. I also considered letting `DecodingExperienceConfig` store a `[String: Any]` dict, but I really don't like building testing-only codepaths into non-test code.

PR #298, opened 2022-12-14
Add leadingFill option for rating components

iujames:
This is the iOS implementation of the [proposed spec change](https://github.com/appcues/appcues-mobile-experience-spec/pull/121/files#diff-b4d74ced281deda59b98f155a0fd068ac0bcaa79d3c9fb67cf91023a80f664f4R457-R460) to add `leadingFill` on the `optionSelect` component.

It is only supported for single select components, with `hidden` controlPosition (ratings use cases)

This would apply in cases like the star rating block being built now.

Screenshot 2022-12-14 at 9 04 39 AM

### DIFF
--- a/Sources/AppcuesKit/Presentation/UI/Components/AppcuesOptionSelect.swift
+++ b/Sources/AppcuesKit/Presentation/UI/Components/AppcuesOptionSelect.swift
@@ -10,6 +10,12 @@ import SwiftUI
 
 @available(iOS 13.0, *)
 internal struct AppcuesOptionSelect: View {
+    private struct OptionItem: Identifiable {
+        var id: String
+        let content: ExperienceComponent
+        let isSelected: Binding<Bool>
+    }
+
     let model: ExperienceComponent.OptionSelectModel
 
     @EnvironmentObject var viewModel: ExperienceStepViewModel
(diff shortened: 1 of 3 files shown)


Sources/AppcuesKit/Presentation/UI/Components/AppcuesOptionSelect.swift
```diff
@@ -55,19 +61,30 @@ internal struct AppcuesOptionSelect: View {
...
+            var content = (isSelected.wrappedValue) ? (item.selectedContent ?? item.content) : item.content
+            if allowLeadingFill,
+                let firstSelectedIndex = firstSelectedIndex,
+                index < firstSelectedIndex,
+                let selectedContent = item.selectedContent {
+                // this overrides an unselected item with selected content styling in the leadingFill case
+                content = selectedContent
             }
+            return OptionItem(id: item.id, content: content, isSelected: isSelected)
+        }
```
iujames, line 84: I rewrote this block a couple of times trying to find a more elegant way to express this concept inside this `@ViewBuilder` block 😅 - open to any ideas to improve or simplify. It just requires the full selection state and sequence of items to apply that leading fill correctly, rather than being focused on each specific item only as before.

PR #299, opened 2022-12-14
Add leadingFill option for rating components

mmaatttt:
An alternative approach to #298


iujames (approved):
nice! pulling out the `.singleLeadingFill` makes this really clear I think.

mmaatttt:
> pulling out the .singleLeadingFill makes this really clear I think
Thanks, yeah the `[String]?` I had initially was enough of a code smell to go that extra step for the sake of clarity.

PR #300, opened 2022-12-15
Add TraitMetadataDelegate to manage properties across trait instances and update backdrop trait to animate

mmaatttt:
We've talked about this change a lot, but here's the summary of specifically where I've landed. It's actually pretty simple, which is definitely a win!

1. A `TraitMetadata` object that stores the current and previous metadata values. The previous values are important to allow certain transitions.
2. Every trait has a `TraitMetadataDelegate` that can be used to:
    1. set a metadata value for reading by other traits
    2. get a metadata value via a handler block. The `type` provided in the block allows a trait to ensure only one instance handles the metadata (rather than, for example, having 2 backdrop traits each trying to animate to the new metadata value). The handler provides a nice wrapper for animating using the metadata values from `@appcues/step-transition-animation`.
    3. Publish a new metadata value on demand. The `TraitComposer` will handle calling publish on a step change, but if a trait knows a change needs to happen otherwise, it can call `publish()` itself
3. The `@appcues/backdrop` trait animates its color change (with a fallback in case the metadata delegate isn't set). I'm intending this to be the general pattern a trait should 
4. The new `@appcues/step-transition-animation` trait. This just provides metadata values that other traits can use to animate. (Note the nice, tidy `init` thanks to `DecodingExperienceConfig`). This is a `ContainerDecoratingTrait` despite not needing the `containerController` reference. I went with this over a new `StepTransitionHandlingTrait` ([ref](https://appcues.slack.com/archives/C031RLGS4F8/p1670527616504029?thread_ts=1670508074.385699&cid=C031RLGS4F8)) that I had previously considered, since this is the only trait that I'm aware of that will not need a reference to either the backdrop or container. In general it feels like an antipattern to have traits that do _nothing_ else, so I don't want to build that as a specific feature of the trait system.

https://user-images.githubusercontent.com/845681/207924236-f8bef97f-b060-4de7-b015-863227625380.mp4

## Next  Steps
1. There's a handful of missing docs on new `public` stuff. Will flesh that out.
2. PRs for the `@appcues/backdrop-keyhole` and `@appcues/target-rectangle`.
3. Create proper specs for all these new traits.

### DIFF
--- a/Sources/AppcuesKit/Presentation/Traits/Appcues/AppcuesStepTransitionAnimationTrait.swift
+++ b/Sources/AppcuesKit/Presentation/Traits/Appcues/AppcuesStepTransitionAnimationTrait.swift
@@ -14,8 +14,8 @@ internal class AppcuesStepTransitionAnimationTrait: ContainerDecoratingTrait {
 
     weak var metadataDelegate: TraitMetadataDelegate?
 
-    let duration: TimeInterval
-    let easing: Easing
+    private let duration: TimeInterval
+    private let easing: Easing
 
     required init?(config: DecodingExperienceConfig, level: ExperienceTraitLevel) {
         self.duration = config["duration"] ?? 0.3
--- a/Sources/AppcuesKit/Presentation/Traits/Appcues/AppcuesStepTransitionAnimationTrait.swift
+++ b/Sources/AppcuesKit/Presentation/Traits/Appcues/AppcuesStepTransitionAnimationTrait.swift
@@ -1,0 +1,68 @@
+//
+//  AppcuesStepTransitionAnimationTrait.swift
+//  AppcuesKit
+//
+//  Created by Matt on 2022-12-06.
+//  Copyright © 2022 Appcues. All rights reserved.
+//
+
+import UIKit
+
+internal class AppcuesStepTransitionAnimationTrait: ContainerDecoratingTrait {
+
+    static var type: String = "@appcues/step-transition-animation"
+
+    weak var metadataDelegate: TraitMetadataDelegate?
+
+    let duration: TimeInterval
+    let easing: Easing
+
+    required init?(config: DecodingExperienceConfig, level: ExperienceTraitLevel) {
+        self.duration = config["duration"] ?? 0.3
+        self.easing = config["easing"] ?? .linear
+    }
+
+    func decorate(containerController: ExperienceContainerViewController) throws {
+        metadataDelegate?.set([
+            "animationDuration": duration,
+            "animationEasing": easing
+        ])
+    }
+
+    func undecorate(containerController: ExperienceContainerViewController) throws {
+        metadataDelegate?.unset(keys: [ "animationDuration", "animationEasing" ])
+    }
+}
+
+extension AppcuesStepTransitionAnimationTrait {
+
+    enum Easing: String, Decodable {
+        case linear, easeIn, easeOut, easeInOut
+
+        var curve: UIView.AnimationOptions {
+            switch self {
+            case .linear:
+                return .curveLinear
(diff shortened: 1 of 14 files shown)


Sources/AppcuesKit/Presentation/Traits/Appcues/AppcuesStepTransitionAnimationTrait.swift
```diff
@@ -0,0 +1,68 @@
...
+import UIKit
+
+internal class AppcuesStepTransitionAnimationTrait: ContainerDecoratingTrait {
+
+    static var type: String = "@appcues/step-transition-animation"
+
+    weak var metadataDelegate: TraitMetadataDelegate?
+
+    let duration: TimeInterval
+    let easing: Easing
```
iujames, line 18: Can these be private?

Sources/AppcuesKit/Presentation/Traits/Appcues/AppcuesStepTransitionAnimationTrait.swift
```diff
@@ -0,0 +1,68 @@
...
+
+    required init?(config: DecodingExperienceConfig, level: ExperienceTraitLevel) {
+        self.duration = config["duration"] ?? 0.3
+        self.easing = config["easing"] ?? .linear
```
iujames, line 22: I still might explore the idea of a trait config instance that gets decoded just to see what that looks like, but this is certainly simple too!

Sources/AppcuesKit/Presentation/Traits/TraitMetadataDelegate.swift
```diff
@@ -0,0 +1,102 @@
...
+import UIKit
+
+/// Methods for managing data to be shared across trait instances.
+@objc
+public class TraitMetadataDelegate: NSObject {
+    private var metadata: [String: Any?] = [:]
+    private var previousMetadata: [String: Any?] = [:]
+
+    private var subscribers: [String: (TraitMetadata) -> Void] = [:]
+    private var viewAnimatingSubscribers: [String: (TraitMetadata) -> Void] = [:]
```
iujames, line 18: on first glance i thought something would get double published if it was in both, maybe could name `subscribers` to `nonAnimatingSubscribers` or something like that - to make it clear these are mutually exclusive sets.

Sources/AppcuesKit/Presentation/Traits/TraitMetadataDelegate.swift
```diff
@@ -0,0 +1,102 @@
...
+        guard let duration = newData["animationDuration"] as? TimeInterval,
+              let easing = newData["animationEasing"] as? AppcuesStepTransitionAnimationTrait.Easing else {
+            return nil
+        }
+
+        let animation = CABasicAnimation(keyPath: keyPath)
+        animation.duration = duration
+        animation.timingFunction = CAMediaTimingFunction(name: easing.timingFunction)
+        return animation
+    }
```
iujames, line 89: Is this one used?

iujames (approved):
Looking good! All seems in line with all the recent convos, to me.

Sources/AppcuesKit/Presentation/Traits/TraitMetadataDelegate.swift
mmaatttt: Good idea!

Sources/AppcuesKit/Presentation/Traits/TraitMetadataDelegate.swift
mmaatttt: Good catch—not yet! It's used by the keyhole trait that will be the next PR

PR #301, opened 2022-12-15
Adjust how border is applied to match CSS border-box style

iujames:
This change reverts and revises what was previously done in #274 , in favor of another approach that will hopefully more closely match the desired outcome of borders being applied in the same way as the CSS border-box style. The issue was around how the border was applied on fixed size elements - and to fix this, the border needs to be applied after the view frame is sized.

When a view does not have any fixed size, just sizing to the intrinsic content size of a label or image, for instance, this inset is _not_ used, and the border just applies on top of the content to increase the overall intrinsic size of the view.

The approach here introduces a new concept of `borderInset`, which gets applied on fixed size height or width elements in a way that reserves some space for a border to be injected later, then sizes the item, then applies the border back into the reserved space to give the desired size with border.

Images and buttons required a little special handling due to how those views are structured. In particular, images using aspectRatio modifiers need to also factor this in when considering whether a view has a fixed size.

The tests and explanation of updates to verify this fix are here https://github.com/appcues/appcues-mobile-experience-spec/pull/123

### DIFF
--- a/Sources/AppcuesKit/Presentation/Extensions/View+Appcues.swift
+++ b/Sources/AppcuesKit/Presentation/Extensions/View+Appcues.swift
@@ -45,11 +45,14 @@ extension View {
     func applyInternalLayout(_ style: AppcuesStyle) -> some View {
         self
             .padding(style.padding)
-            .applyBorderStyle(style)
             .frame(width: style.width, height: style.height)
             .if(style.fillWidth) { view in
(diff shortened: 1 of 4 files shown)


mmaatttt (approved):
Wow, really nice! We should probably add more snapshot test cases with different combos of fixed width and/or height on the different components to stress test the `borderInset`

PR #302, opened 2022-12-16
Add keyhole trait

mmaatttt:
Nice and simple! The only complicated bit is figuring out the keyhole rectangle coordinates where we need to handle 3 cases: nothing->keyhole, keyhole->keyhole, keyhole->nothing.

https://user-images.githubusercontent.com/845681/208162941-8c6c8cda-a366-4dd4-a10f-0aa92143c5b7.mp4


iujames (approved):
looks great!

PR #303, opened 2022-12-19
Feature/trait init

iujames:
note: this is based on `feature/sc-44387/keyhole-trait` for now, since I wanted to include all the latest changes.

This PR is a fairly simple refactor that ended up touching a fair bit of code in the main SDK and then a LOT of code in the tests to re-work accordingly. It doesn't really add anything new, but was an idea from discussion that I wanted to try out and see how it looked, and we can decide if worthwhile or not.

The core idea
* Traits and Actions are initialized now with a type called `ExperiencePluginConfiguration`
* Traits and Actions defined their own config structs (Decodable) and call `configuration.decode(Config.self)` to fetch config values
* There is no longer a concept of a partially decoded config - each "Plugin" (naming open for change - Action or Trait) is responsible for defining it's own JSON contract
* Internally, a protocol `PluginDecoder` is used to inject decoding functionality int `Experience.Trait` and `Experience.Action`, during decoding of those items - and implementations capture the KeyedDecodingContainer to use accordingly. This is hidden away inside those objects, but this protocol approach allows for fairly easily mocking this in tests

There were several other spots in tests where some initializer extensions were used to simplify syntax

In the end, I think it's a moderate improvement in terms of the init structure for these - providing a way to have strongly typed config objects, no string keys floating around, and hopefully more maintainable tests. It ended up being more code changed that I was expecting when first trying the idea though.


Sources/AppcuesKit/Presentation/Public/ExperiencePluginConfiguration.swift
```diff
@@ -0,0 +1,23 @@
...
+    private var decoder: PluginDecoder
+
+    init(_ decoder: PluginDecoder) {
+        self.decoder = decoder
+    }
+
+    public func decode<T: Decodable>(_ type: T.Type) -> T? {
+        return decoder.decode(type)
+    }
+}
```
iujames, line 23: the new init object, with the `PluginDecoder` injection that allows it to be used across Actions and Traits, as well as more easily tested with mock decoder.

Tests/AppcuesKitTests/Actions/AppcuesContinueActionTests.swift
```diff
@@ -59,3 +59,16 @@ class AppcuesContinueActionTests: XCTestCase {
...
+    convenience init?() {
+        self.init(configuration: ExperiencePluginConfiguration(nil))
+    }
+    convenience init?(index: Int? = nil, offset: Int? = nil, stepID: String? = nil) {
+        var uuid: UUID?
+        if let stepID = stepID {
+            uuid = UUID(uuidString: stepID)
+        }
+        self.init(configuration: ExperiencePluginConfiguration(AppcuesContinueAction.Config(index: index, offset: offset, stepID: uuid)))
+    }
```
iujames, line 73: made various convenience inits on Actions and Traits like this in tests to just encapsulate the usage of the underlying Config type and simplify testing callsites.

Sources/AppcuesKit/Presentation/Actions/Appcues/AppcuesUpdateProfileAction.swift
```diff
@@ -29,3 +33,32 @@ internal class AppcuesUpdateProfileAction: ExperienceAction {
         completion()
     }
 }
+
+@available(iOS 13.0, *)
+extension AppcuesUpdateProfileAction.Config: Decodable {
```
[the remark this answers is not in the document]
mmaatttt: > internal to this one usage spot now

That's a nice improvement!

Sources/AppcuesKit/Presentation/Traits/Appcues/AppcuesStepTransitionAnimationTrait.swift
```diff
@@ -17,9 +21,10 @@ internal class AppcuesStepTransitionAnimationTrait: ContainerDecoratingTrait {
     private let duration: TimeInterval
     private let easing: Easing
 
-    required init?(config: DecodingExperienceConfig, level: ExperienceTraitLevel) {
-        self.duration = config["duration"] ?? 0.3
-        self.easing = config["easing"] ?? .linear
+    required init?(configuration: ExperiencePluginConfiguration, level: ExperienceTraitLevel) {
+        let config = configuration.decode(Config.self)
```
mmaatttt, line 25: I like the consistency of `guard let config = configuration.decode(Config.self)`  in every trait & action init, even if the decode will never fail (like in this case).

Sources/AppcuesKit/Presentation/Traits/Appcues/AppcuesStepTransitionAnimationTrait.swift
```diff
@@ -9,6 +9,10 @@
 import UIKit
 
 internal class AppcuesStepTransitionAnimationTrait: ContainerDecoratingTrait {
+    struct Config: Decodable {
+        let duration: Double?
+        let easing: String?
```
mmaatttt, line 14: I think the type here should be able to be `Easing?`, since that's also `Decodable`.

Tests/AppcuesKitTests/ExperiencePluginConfiguration+Testable.swift
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    func decode<T>(_ type: T.Type) -> T? where T : Decodable {
+        guard let value = value else { return nil }
+        guard let castValue = value as? T else {
+            fatalError("uh oh \(type): \(value)")
+        }
+        return castValue
+    }
+}
```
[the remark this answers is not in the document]
mmaatttt: Clever!

Tests/AppcuesKitTests/Actions/AppcuesContinueActionTests.swift
mmaatttt: I'm not sure I agree with doing this everywhere (specifically in the various `testInit()` tests. Reason being in these tests we're explicitly asserting that, for example, a `nil` init does _whatever_, but adding the layer of indirection with a testing-only convenience init makes it harder to understand what the test is actually testing. For these cases, I'd prefer to keep `AppcuesContinueAction(configuration: ExperiencePluginConfiguration(nil))` or whatever.

Absolutely makes sense for test-only actions/traits and any tests that aren't specifically testing the initializer.

mmaatttt: I'm not sure I agree with doing this everywhere (specifically in the various `testInit()` tests. Reason being in these tests we're explicitly asserting that, for example, a `nil` init does _whatever_, but adding the layer of indirection with a testing-only convenience init makes it harder to understand what the test is actually testing. For these cases, I'd prefer to keep `AppcuesContinueAction(configuration: ExperiencePluginConfiguration(nil))` or whatever.

Absolutely makes sense for test-only actions/traits and any tests that aren't specifically testing the initializer.

Tests/AppcuesKitTests/Traits/TraitComposerTests.swift
```diff
@@ -459,6 +458,48 @@ extension TraitComposerTests {
...
+                 containerCreatingExpectation: XCTestExpectation? = nil,
+                 containerDecoratingExpectation: XCTestExpectation? = nil,
+                 wrapperCreatingExpectation: XCTestExpectation? = nil,
+                 backdropDecoratingExpectation: XCTestExpectation?) {
+                self.groupID = groupID
+                if let stepDecoratingExpectation = stepDecoratingExpectation {
+                    self.stepDecoratingExpectation = DecodableExpectation(expectation: stepDecoratingExpectation)
+                } else {
+                    self.stepDecoratingExpectation = nil
+                }
```
mmaatttt, line 480: Nothing wrong with this (verbosity is fine in tests), but if you want to make it a 1-liner you could make the init accept an optional expectation and then return `nil` if that's what you get.

Sources/AppcuesKit/Presentation/Traits/Appcues/AppcuesStepTransitionAnimationTrait.swift
iujames: I _think_ we can do that in most or all. I just don't want to accidentally fail something in the case of: config in it's entirety is optional (can go with all defaults), but if it is not-omitted, then it has some non-optional values. Need to make sure all the cases are covered to let it gracefully fall back to defaults in such cases. Need to review all the usages for that.

Tests/AppcuesKitTests/Traits/TraitComposerTests.swift
iujames: right, thank you - happy to compress this.

Sources/AppcuesKit/Presentation/Traits/Appcues/AppcuesStepTransitionAnimationTrait.swift
iujames: After some discussion, leaving as-is for now.  This way the config can capture the required/optional json properties. It is up to the trait/action implementor to declare if completely missing config is ok or not, for that particular case.

PR #304, opened 2022-12-20
Add skeleton @appcues/tooltip trait

mmaatttt:
There's a lot of similarities between the standard modal and a tooltip container, so I've reworked `DialogContainerViewController` and `DialogContainerView` to be a superclass for both.

The tooltip trait here doesn't do much yet (ie no pointer), but it handles the case with no target element and otherwise it always positions the wrapper 20pt below the target rectangle:

|With target rectangle|Without target rectangle|
|-|-|
|Simulator Screen Shot - iPhone 14 Pro - 2022-12-20 at 09 27 19|Simulator Screen Shot - iPhone 14 Pro - 2022-12-20 at 09 28 04|

Obvious _lots_ more to do, but wanted to start with a small PR to make the reworking Dialog bit clearer to follow.


iujames (approved):
Nice call consolidating and sharing this code - that should lead to some nice efficiencies for some of the ideas coming up. Side note - the inline codecov warnings here make these diffs a bit hard to read somtimes, is there a way to hide those?

mmaatttt:
@iujames on a file-by-file basis, you can select the 3 dots in the top right and hide the annotations. Alternatively I should be able to update the codecov config if we don't find it useful at all (I really don't)

```
github_checks:
    annotations: false
```

PR #306, opened 2023-01-03
Update form components to handle label horizontalAlignment

mmaatttt:
Updating the option select **(edit: and text)** component layout to:
1. Always take the full width available (the `HStack` causes this)
2. Visually apply the `optionSelect.style.horizontalAlignment` to the options (L22, but it didn't change)
3. Apply the `optionSelect.label.style.horizontalAlignment` to the question label (previous it was ignored, will default to leading if no value is specified)
4. Apply the `optionSelect.errorLabel.style.horizontalAlignment` to the error label (previous it was ignored, will default to leading if no value is specified)

Consequently, the question label, options, and error can all be horizontally aligned independently of each other.

|Fixed|Broken|
|-|-|
|Simulator Screen Shot - iPhone 14 Pro - 2023-01-03 at 13 03 43|Simulator Screen Shot - iPhone 14 Pro - 2023-01-03 at 13 08 11|

I also fixed a potential SwiftUI error I noticed when copy/pasting stuff around an example experience: an empty equal width horizontal stack would try to divide by 0. SwiftUI is smart enough not to crash ("Invalid frame dimension (negative or non-finite)"), but it would case an undefined behaviour.

### DIFF
--- a/Sources/AppcuesKit/Presentation/UI/Components/AppcuesOptionSelect.swift
+++ b/Sources/AppcuesKit/Presentation/UI/Components/AppcuesOptionSelect.swift
@@ -20,7 +20,16 @@ internal struct AppcuesOptionSelect: View {
         let errorTintColor = stepState.shouldShowError(for: model.id) ? Color(dynamicColor: model.errorLabel?.style?.foregroundColor) : nil
 
         VStack(alignment: style.horizontalAlignment, spacing: 0) {
-            TintedTextView(model: model.label, tintColor: errorTintColor)
+            HStack {
+                // nil-coalesce to .leading so a non-specified value defaults to leading-aligned
+                if HorizontalAlignment(string: model.label.style?.horizontalAlignment) ?? .leading != .leading {
+                    Spacer()
+                }
+                TintedTextView(model: model.label, tintColor: errorTintColor)
+                if HorizontalAlignment(string: model.label.style?.horizontalAlignment) != .trailing {
+                    Spacer()
+                }
+            }
 
             switch (model.selectMode, model.displayFormat) {
             case (.single, .picker):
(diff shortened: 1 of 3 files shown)


Sources/AppcuesKit/Presentation/UI/Components/AppcuesOptionSelect.swift
```diff
@@ -20,7 +20,16 @@ internal struct AppcuesOptionSelect: View {
...
+            HStack {
+                // nil-coalesce to .leading so a non-specified value defaults to leading-aligned
+                if HorizontalAlignment(string: model.label.style?.horizontalAlignment) ?? .leading != .leading {
+                    Spacer()
+                }
+                TintedTextView(model: model.label, tintColor: errorTintColor)
+                if HorizontalAlignment(string: model.label.style?.horizontalAlignment) != .trailing {
+                    Spacer()
+                }
+            }
```
iujames, line 32: only other approach that comes to mind, and what I'm considering on the Android side - using nested `VStack` set to fill parent width, and then enforcing their own horizontal alignment within. This seems fine too.

iujames (approved):

Sources/AppcuesKit/Presentation/UI/Components/AppcuesOptionSelect.swift
mmaatttt: A nested `VStack` would be nice, but actually doesn't work (at least simply) because frame on the stack doesn't affect the layout within:
Screenshot 2023-01-03 at 1 35 58 PM

iujames: oh, darn, that seems kinda weird. On Compose it does fill the parent width, and then align the nested contents within that full width.

PR #309, opened 2023-01-05
Update swift and ruby dependencies

mmaatttt:
Also fixing a few old and new lint warnings.


.swiftlint.yml
```diff
@@ -4,7 +4,6 @@ disabled_rules:
   -  no_fallthrough_only
 
 opt_in_rules:
-  - anyobject_protocol
```
mmaatttt, line 7: This is now handled by the Swift compiler, so SwiftLint has deprecated it

mmaatttt:
@iujames the new mint packages are properly cached and the CI run time is back to normal:
Screenshot 2023-01-05 at 4 22 44 PM

iujames (approved):

PR #310, opened 2023-01-06
Fix swift package compilation issue with tests reference iOS 13+ types

iujames:
Should patch up the build issue reported here https://swiftpackageindex.com/appcues/appcues-ios-sdk/builds, and restore our badge display.

### DIFF
--- a/Tests/AppcuesKitTests/Analytics/AnalyticsTrackerTests.swift
+++ b/Tests/AppcuesKitTests/Analytics/AnalyticsTrackerTests.swift
@@ -98,24 +98,10 @@ extension Dictionary where Key == String, Value == Any {
         }
         XCTAssertEqual(Set(self.keys), Set(other.keys), file: file, line: line)
         self.keys.forEach { key in
-            if #available(iOS 13.0, *) {
-                // can reference ExperienceData.StepState on 13+
-                switch(self[key], other[key]) {
-                case let (val1 as String, val2 as String):
-                    XCTAssertEqual(val1, val2, file: file, line: line)
-                case let (val1 as NSNumber, val2 as NSNumber):
-                    XCTAssertEqual(val1, val2, file: file, line: line)
-                case let (val1 as [Any], val2 as [Any]):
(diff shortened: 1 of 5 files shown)


Tests/AppcuesKitTests/Analytics/AnalyticsTrackerTests.swift
```diff
@@ -98,19 +98,36 @@ extension Dictionary where Key == String, Value == Any {
...
-            case let (val1 as [Any], val2 as [Any]):
-                val1.verifyPropertiesMatch(val2, file: file, line: line)
-            case let (val1 as [String: Any], val2 as [String: Any]):
-                val1.verifyPropertiesMatch(val2, file: file, line: line)
-            case let (val1 as ExperienceData.StepState, val2 as ExperienceData.StepState):
-                XCTAssertEqual(val1, val2, file: file, line: line)
-            default:
-                XCTFail("\(self[key] ?? "nil") does not match \(other[key] ?? "nil").", file: file, line: line)
+            if #available(iOS 13.0, *) {
+                // can reference ExperienceData.StepState on 13+
```
iujames, line 102: wasn't sure best route on this one (and one below). If the entire extension or function is marked 13+, a dozen or so other test callsites would need updated, most of which not having anything to do with `ExperienceData.StepState`. This approach isolates the change to these spots, but results in some redundancy in the checks within. If we don't care about just marking various other tests 13+ only even though they are not really related to presentation stuff, can do that too.

mmaatttt: There's a less repetitive approach we've used elsewhere that might work nicely here?

https://github.com/appcues/appcues-ios-sdk/blob/main/Sources/AppcuesKit/Data/Networking/DynamicCodingKeys.swift#L45-L49

iujames: yes, thank you. https://github.com/appcues/appcues-ios-sdk/pull/310/commits/eb3af4ad8c618c31ab05f3c4ffd344a785745bcb

PR #311, opened 2023-01-11
Add new deep link route that launches the debugger in screen capture mode

iujames:
(note, currently pointing this at `traits-2.0` since it's related to the element targeting work, though technically independent of 2.0 updates and hidden behind deep link access only. I don't see any reason why it would need to be in an earlier release though at this point.)

This adds support for a new deep link path `appcues-{app_id}://sdk/capture_screen`

This path will launch the UIDebugger with a `DebugMode` parameter (new). There are now two modes: `.debugger` - everything we had previously and `.screenCapture` - the new mode for capturing snapshot and layout hierarchy. 

This approach was intended to maximize re-use of existing Debugger framework and just skin it with a different icon and click action. This worked out fairly reasonably, and kept the majority of the changes in the `FloatingView` and associated file. A new `FloatingImageView` was introduced, which is really just the contents within the `FloatingView` - since the previous implementation used an image asset with a background gradient baked in, and this new mode uses a vector image laid over top of a blurple haze gradient. We may want to reconcile this to a common approach.

A couple other touch points
* don't send the floating log messages to the debugger unless in `.debugger` mode
* if a new request to show the debugger comes in, check the mode - if it changes, we'll get rid of the current window and launch a new window in the new mode. Previously, this just skipped out early since it considered the debugger already open.
* reworked how the full debug panel gets initialized (the `DebugUI.MainPanelView`) slightly - moving that into `DebugViewController` and only creating it when the mode is `.debugger`

This PR does not include any work on the actual screen capture itself, just the framework to launch this mode - screen capture behaviors would be coming next.

Screenshot 2023-01-11 at 1 41 15 PM


Sources/AppcuesKit/Presentation/Debugger/FloatingView/DebugViewController.swift
```diff
@@ -16,12 +17,14 @@ internal class DebugViewController: UIViewController {
...
-    private var debugView = DebugView()
+    private lazy var debugView = DebugView(mode: mode)
 
-    private let panelViewController: UIViewController
+    let mode: DebugMode
+    private let viewModel: DebugViewModel
 
-    init(wrapping panelViewController: UIViewController) {
-        self.panelViewController = panelViewController
+    init(viewModel: DebugViewModel, mode: DebugMode) {
```
mmaatttt, line 25: Does this handle the case where the debugger is loaded in one mode and then the other is tried? Like if I have the regular debugger open and then try the capture deep link? Everything looks tied to the init as far as I can see.

Sources/AppcuesKit/Presentation/Debugger/FloatingView/FloatingView.swift
```diff
@@ -169,3 +211,39 @@ internal class FloatingView: UIView {
...
+    }
+
+    var imageView: UIView {
+        switch self {
+        case .debugger:
+            let imageView = UIImageView(image: UIImage(asset: Asset.Image.debugIcon))
+            imageView.backgroundColor = .secondarySystemBackground
+            return imageView
+        case.screenCapture:
+            return FloatingImageView(image: UIImage(asset: Asset.Image.captureScreen), size: CGSize(width: 24.0, height: 24.0))
```
mmaatttt, line 233: My preference would be to update the asset to include the gradient for the sake of simplicity, but this certainly works.

Sources/AppcuesKit/Presentation/Debugger/FloatingView/DebugView.swift
```diff
@@ -11,6 +11,7 @@ import UIKit
 @available(iOS 13.0, *)
 internal protocol DebugViewDelegate: AnyObject {
     func debugView(did event: DebugView.Event)
+    func screenCaptured()
```
mmaatttt, line 14: This is fine as is, but another option would be to add `screenCapture` as a case to `DebugView.Event`.

Sources/AppcuesKit/Presentation/Debugger/FloatingView/DebugView.swift
```diff
@@ -124,8 +127,9 @@ internal class DebugView: UIView, FloatingViewDelegate {
 
     // MARK: Overrides
 
-    override init(frame: CGRect) {
-        super.init(frame: frame)
+    init(mode: DebugMode) {
```
mmaatttt, line 130: I might try and make it where `DebugView` is entirely agnostic to `DebugMode`. Maybe have the `DebugViewController` be the `FloatingViewDelegate` instead of this (that's my bad from how it was), and then the VC could also toggle the icon on the floating view?

Seems less than ideal to handle interactions and state in the view like this (again, my bad). It's also

mmaatttt (approved):
I think this all fits in well with the way the debugger is currently architected. The debugger architecture could definitely be better, but these changes won't make it any more difficult to clean that up if/when we feel its necessary.

Sources/AppcuesKit/Presentation/Debugger/FloatingView/DebugViewController.swift
iujames: That part is currently handled a level up, in `UIDebugger.show(mode: DebugMode)` -- it will hide the old debugger instance and create a new one entirely when switching modes.

mmaatttt: Gotcha. I guess it makes sense to just nuke it and start over since we want the floating button, etc to get reset when launching the new mode.

Sources/AppcuesKit/Presentation/Debugger/FloatingView/DebugView.swift
iujames: Good call, making consistent with others.

Sources/AppcuesKit/Presentation/Debugger/FloatingView/DebugView.swift
iujames: This worked out nicely I think, simpler - thanks

PR #312, opened 2023-01-12
Support relative positions in @appcues/target-rectangle trait

mmaatttt:
Turns out to be quite straightforward. We need to watch the backdrop bounds and then recalculate the position when they change. The `@appcues/target-element` trait will do the exact same thing.

The subview is a simple way to monitor the bounds, but definitely open to a cleaner solution there.


iujames (approved):
nice. I can't really think of any other way to track the bounds of the backdrop. The risk here being that some other trait could theoretically manipulate the view hierarchy and remove this view I suppose. But that would be pretty reckless trait behavior.

PR #314, opened 2023-01-17
Add support for feathering to @appcues/backdrop-keyhole trait

mmaatttt:
If it's a rectangle we use the `CAShapeLayer` approach from before. A circle (feathering or not) uses the new `CAGradientLayer`. A circle without feathering could use the `CAShapeLayer` still but the animation between circle with feathering to circle without feathering would be janky or a bunch more code to support.

The tricky part is that the `CAGradientLayer` startPoint/endPoint/locations are all relative, so it took a bit of thought to calculate properly.

If we're really gunning for pixel perfect, we'd need a 2 step animation from a rectangle to a circle with feather since right now there's a slight jump.

|Small Radius|Large Radius|
|-|-|
|Simulator Screen Shot - iPhone 14 Pro - 2023-01-17 at 11 50 57|Simulator Screen Shot - iPhone 14 Pro - 2023-01-17 at 11 50 51|


Sources/AppcuesKit/Presentation/Traits/Appcues/AppcuesBackdropKeyholeTrait.swift
```diff
@@ -97,15 +98,85 @@ internal class AppcuesBackdropKeyholeTrait: BackdropDecoratingTrait {
...
+            let newStartPoint = CGPoint(
+                x: newKeyholeBezierPath.bounds.midX,
+                y: newKeyholeBezierPath.bounds.midY)
+            let newEndPoint = CGPoint(
+                x: newKeyholeBezierPath.bounds.maxX + newBlurRadius,
+                y: newKeyholeBezierPath.bounds.maxY + newBlurRadius)
+
+            // Start the gradient at the edge of the keyhole path.
+            // Doesn't matter whether we use x or y since it's a circle.
+            // Cap at 0.995 so we don't collide with the
```
iujames, line 113: > so we don't collide with the

looks like the end of sentence may have gotten dropped

Sources/AppcuesKit/Presentation/Traits/Appcues/AppcuesBackdropKeyholeTrait.swift
```diff
@@ -97,15 +98,85 @@ internal class AppcuesBackdropKeyholeTrait: BackdropDecoratingTrait {
...
+            let newStartLocation: Double = min(
+                0.995,
+                (newEndPoint.x - newBlurRadius - newStartPoint.x) / (newEndPoint.x - newStartPoint.x)
```
iujames, line 116: is there any chance of div/0 here?

Sources/AppcuesKit/Presentation/Traits/Appcues/AppcuesBackdropKeyholeTrait.swift
```diff
@@ -97,15 +98,85 @@ internal class AppcuesBackdropKeyholeTrait: BackdropDecoratingTrait {
...
+                }()
+                let oldStartPoint = CGPoint(
+                    x: oldKeyholeBezierPath.bounds.midX,
+                    y: oldKeyholeBezierPath.bounds.midY)
+                let oldEndPoint = CGPoint(
+                    x: oldKeyholeBezierPath.bounds.maxX + oldBlurRadius,
+                    y: oldKeyholeBezierPath.bounds.maxY + oldBlurRadius)
+                let oldStartLocation: Double = min(
+                    0.995,
+                    (oldEndPoint.x - oldBlurRadius - oldStartPoint.x) / (oldEndPoint.x - oldStartPoint.x)
```
iujames, line 144: or here?

iujames (approved):
wizardry.  looks great!

Sources/AppcuesKit/Presentation/Traits/Appcues/AppcuesBackdropKeyholeTrait.swift
mmaatttt: It would require the blur radius to be 0 and the width of the target to also be 0, but possible in theory, so I've added a check.

PR #315, opened 2023-01-17
Add confirmation dialog and toast for screen capture flow

iujames:
stacks on #313 

The UI side of things for the capture flow shown in the video below. Did this stuff in SwiftUI just because why not. It doesn't necessarily have to be though. This hooks in to the DebugViewController to present the modal, and then get the updated screen name and pass along for upcoming API integration work (TBD).

https://user-images.githubusercontent.com/19266448/212973639-9d826f29-cf35-4fb7-b82c-70f04cd83e68.mov

### DIFF
--- a/Sources/AppcuesKit/Presentation/Debugger/FloatingView/DebugView.swift
+++ b/Sources/AppcuesKit/Presentation/Debugger/FloatingView/DebugView.swift
@@ -73,6 +73,14 @@ internal class DebugView: UIView {
         return view
     }()
 
+    var toastWrapperView: UIView = {
+        let view = UIView()
+        view.translatesAutoresizingMaskIntoConstraints = false
+        view.alpha = 0
+
+        return view
+    }()
+
     // MARK: Layout Calculators
 
     var baseFrame: CGRect = .zero
@@ -131,6 +139,7 @@ internal class DebugView: UIView {
         addSubview(panelWrapperView)
         addSubview(dismissView)
         addSubview(floatingView)
+        addSubview(toastWrapperView)
 
         backgroundView.pin(to: self)
         NSLayoutConstraint.activate([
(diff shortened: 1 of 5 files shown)


Sources/AppcuesKit/Presentation/Debugger/FloatingView/DebugView.swift
```diff
@@ -441,6 +467,23 @@ internal class DebugView: UIView {
...
+
+        if animated {
+            UIView.animate(
+                withDuration: 0.6,
+                animations: animations) { _ in completion?() }
+        } else {
+            animations()
+            completion?()
+        }
+    }
```
iujames, line 485: If we ever have a more general need for toast capabilities, perhaps this functionality gets hoisted up elsewhere and generalized. For now, I decided to keep it localized and specific to the one and only known use case here.

Sources/AppcuesKit/Presentation/Debugger/FloatingView/DebugViewController.swift
```diff
@@ -57,8 +57,8 @@ internal class DebugViewController: UIViewController {
         debugView.setFloatingView(visible: true, animated: animated, programmatically: true)
     }
 
-    func hide(animated: Bool) {
-        debugView.setFloatingView(visible: false, animated: animated, programmatically: true)
+    func hide(animated: Bool, notify: Bool = false) {
+        debugView.setFloatingView(visible: false, animated: animated, programmatically: true, notify: notify)
```
iujames, line 61: the `notify` is so we can hide the FAB while the confirmation model is up, and that hiding does not get interpreted as the user closing the debugger entirely, as a normal hide action would do.

Sources/AppcuesKit/Presentation/Debugger/FloatingView/DebugView.swift
```diff
@@ -179,6 +193,12 @@ internal class DebugView: UIView {
     }
 
     override func hitTest(_ point: CGPoint, with event: UIEvent?) -> UIView? {
+
+        // any interaction while toast is showing will initiate dismiss
+        if toastWrapperView.alpha > 0 {
+            animateToast(visible: false, animated: true, completion: nil)
+        }
```
mmaatttt, line 200: This is very clever! Almost feels too clever but it'd be a bunch more code to do a different way.

mmaatttt (approved):

PR #316, opened 2023-01-17
Add AppcuesTargetElementTrait

iujames:
stacks on #313 

Adds support for `@appcues/target-element` -- this one is very similar to the target rectangle, but adds support for searching through the layout tree to find a matching element based on an `ElementSelector` from the trait config. This trait can throw, if no match is found, or if multiple non-distinct matches are found.

example usage:
```js
"traits": [
	{
		"type": "@appcues/target-element",
		"config": {
			"selector": {
				"accessibilityIdentifier": "tabProfile"
			}	
		}
	},
	{
		"type": "@appcues/backdrop-keyhole",
		"config": {
			"shape": "circle"
		}
	}
]
```

just showing the keyhole here with a modal flow, since I don't have all the latest tooltip presentation code locally
Screenshot 2023-01-17 at 5 04 37 PM


Sources/AppcuesKit/Presentation/Traits/Appcues/AppcuesTargetElementTrait.swift
```diff
@@ -0,0 +1,142 @@
...
+        guard let criteria = criteria else { return nil }
+        let matches = self.filter {
+            guard let selector = $0.appcuesSelector else { return false }
+            return accessor(selector) == criteria
+        }
+        if matches.count == 1 {
+            return matches[0]
+        }
+        return nil
+    }
```
mmaatttt, line 141: I think you can use a KeyPath to do this a bit more elegantly:
```swift
private extension Array where Element == UIView {
    // find a single item that matches the criteria, based on the accessor
    // i.e. a single item that matches the accessibilityIdentifier
    func findSingleMatch(for targetSelector: ElementSelector, on keyPath: KeyPath<ElementSelector, String?>) -> UIView? {
        let matches = self.filter {
            guard let selector = $0.appcuesSelector else { return false }
            return selector[keyPath: keyPath] == targetSelector[keyPath: keyPath]
        }
        if matches.count == 1 {
            return matches[0]
        }
        return nil
    }
}

// usage
views.findSingleMatch(for: target, on: \.accessibilityIdentifier)
views.findSingleMatch(for: target, on: \.tag)
views.findSingleMatch(for: target, on: \.description)
```

PR #318, opened 2023-01-18
Apply traits after any pre-step container navigation executes

iujames:
This is an update to support the use case of: pre-step navigation, then show a step with a targeted-element that depends on the navigation to the new screen being completed.

Previously, the call to the `TraitComposer.package` function would result in the backdrop decorating traits applying their initial decorations - which could result in target-element lookups occurring. This was happening prior to the execution of the navigation actions, before the presentation of the new container.

Now, the application of initial step decoration is delayed until later in the process, after navigation actions have had a chance to execute - in hte `executePresentContainer` function of `ExperienceStateMachine.SideEffect`

note: this is targeting `traits-2.0` as it primarily relates to target-element behavior. Technically, we could port back a similar (but different) update to the 1.x branch which _may_ be applicable to navigation step usage prior to this update, but it is more likely this will come up as an issue in 2.0 targeted element flows. The trait application flow is significantly different in the 1.x branch.

### DIFF
--- a/Tests/AppcuesKitTests/Traits/TraitComposerTests.swift
+++ b/Tests/AppcuesKitTests/Traits/TraitComposerTests.swift
@@ -56,9 +56,12 @@ class TraitComposerTests: XCTestCase {
             ],
             redirectURL: nil,
             nextContentID: nil)
+        let stepIndex = Experience.StepIndex(group: 0, item: 0)
+        let package = try traitComposer.package(experience: ExperienceData(experience, trigger: .showCall), stepIndex: stepIndex)
 
         // Act
-        _ = try traitComposer.package(experience: ExperienceData(experience, trigger: .showCall), stepIndex: Experience.StepIndex(group: 0, item: 0))
+        // simulate the presentation, this updater called from ExperienceStateMachine.SideEffect.executePresentContainer
+        try package.stepDecoratingTraitUpdater(stepIndex.item, nil)
 
         // Assert
         waitForExpectations(timeout: 1)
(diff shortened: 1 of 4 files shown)


mmaatttt (approved):
This ended up being a pretty simple change! I'm always impressed how flexible the trait and state machine stuff ends up being. It can be dense to get to the right part, but it's super flexible and powerful.

Good note on the `traits-2.0` differences. I agree we don't have a need for a similar change on the 1.x releases.

Tests/AppcuesKitTests/Traits/TraitComposerTests.swift
```diff
@@ -56,9 +56,12 @@ class TraitComposerTests: XCTestCase {
             ],
             redirectURL: nil,
             nextContentID: nil)
+        let stepIndex = Experience.StepIndex(group: 0, item: 0)
+        let package = try traitComposer.package(experience: ExperienceData(experience, trigger: .showCall), stepIndex: stepIndex)
 
         // Act
-        _ = try traitComposer.package(experience: ExperienceData(experience, trigger: .showCall), stepIndex: Experience.StepIndex(group: 0, item: 0))
+        // simulate the presentation, this updater called from ExperienceStateMachine.SideEffect.executePresentContainer
+        try package.stepDecoratingTraitUpdater(stepIndex.item, nil)
```
iujames, line 64: fyi @mmaatttt on these test changes to match - I had missed these earlier. I believe the primary intent of these is to  test trait propagation, and this update will preserve the same underlying test - though it's now triggered in a different fashion.

mmaatttt: Yep, perfect

PR #319, opened 2023-01-19
Update SwiftLint identifier_name rule to allow x and y

iujames:
follow up item from https://github.com/appcues/appcues-ios-sdk/pull/313#discussion_r1072605331


mmaatttt (approved):
Thank you!

PR #320, opened 2023-01-19
Fix double publishing of the target rectangle frame

mmaatttt:
The target rectangle trait was doing an extra metadata publish on step change from within the frame observer.

Publishing twice on a step change breaks the `@appcues/backdrop-keyhole` animation because the first publish has the correct values to animate but that animation gets replaced by the second publish which has the same values.

Calling `layoutIfNeeded()` *before* adding the observer takes care of the double. Then there's an edge case for the first step in a group where the backdrop view still has zero bounds because it's not in the view hierarchy yet, so checking for the `window` being set means we can skip publishing a value the first time and get a proper animation triggered by the frame observer.

The `@appcues/target-element` trait will likely require similar changes, but I'll tackle that separately for the demo.

### DIFF
--- a/Sources/AppcuesKit/Presentation/Traits/Appcues/AppcuesTargetRectangleTrait.swift
+++ b/Sources/AppcuesKit/Presentation/Traits/Appcues/AppcuesTargetRectangleTrait.swift
@@ -37,11 +37,17 @@ internal class AppcuesTargetRectangleTrait: BackdropDecoratingTrait {
     }
 
     func decorate(backdropView: UIView) throws {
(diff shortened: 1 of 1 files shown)


iujames (approved):

iujames:
confirmed a similar update on `AppcuesTargetElementTrait` allows animation support on keyhole in my test flow 👍

mmaatttt:
@iujames sweet, if you can PR that, you'll save me 10 min of demo prep

iujames:
sure will do

PR #321, opened 2023-01-19
Update layout capture structure

iujames:
This updates the structure to match recent [discussions](https://appcues.slack.com/archives/C02EX9N4067/p1674141246499829) with Humber around API contract. Specifically, moving some fields inside of an unstructured `metadata` property they will store, for future flexibility.


mmaatttt (approved):
Bonus win makes encoding less verbose 🙌

PR #322, opened 2023-01-19
Fix double publishing of target element frame

iujames:
The target element equivalent of #320

### DIFF
--- a/Sources/AppcuesKit/Presentation/Traits/Appcues/AppcuesTargetElementTrait.swift
+++ b/Sources/AppcuesKit/Presentation/Traits/Appcues/AppcuesTargetElementTrait.swift
@@ -16,12 +16,12 @@ internal class AppcuesTargetElementTrait: BackdropDecoratingTrait {
 
     private class FrameObserverView: UIView {
         private var oldBounds: CGRect = .zero
-        var onChange: ((_ bounds: CGRect) -> Void)?
+        var onChange: (() -> Void)?
 
         override func layoutSubviews() {
             super.layoutSubviews()
             if oldBounds != bounds {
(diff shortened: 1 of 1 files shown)


Sources/AppcuesKit/Presentation/Traits/Appcues/AppcuesTargetElementTrait.swift
```diff
@@ -41,11 +41,19 @@ internal class AppcuesTargetElementTrait: BackdropDecoratingTrait {
     }
 
     func decorate(backdropView: UIView) throws {
-        metadataDelegate?.set([ "targetRectangle": try calculateRect(bounds: backdropView.bounds) ])
+        // Determine if the selector can be resolved to a view, failing the step if not.
+        let targetRect = try calculateRect(bounds: backdropView.bounds)
```
iujames, line 45: subtle difference here, at least for now - since we want to fail on bad selector, we need to still check for the matching view here.  Goes back to the convo on error handling in the frame observer, which has not been resolved yet.

mmaatttt (approved):

Sources/AppcuesKit/Presentation/Traits/Appcues/AppcuesTargetElementTrait.swift
mmaatttt: Makes sense. Also, it doesn't look the `bounds: backdropView.bounds` is used anywhere.

iujames: ah you are right. will clean up. thanks

PR #324, opened 2023-01-19
Add accessibility identifiers to Cocoapods example app

iujames:
This matches the identifiers that were previously updated in the SPM example in #317, so we can test target element behaviors with this app.

events tab
* tabEvents
* btnEvent1
* btnEvent2

profile tab
* tabProfile
* txtGivenName
* txtFamilyName
* btnSaveProfile

group tab
* tabGroup
* txtGroup
* btnSaveGroup

### DIFF
--- a/Examples/DeveloperCocoapodsExample/Podfile.lock
+++ b/Examples/DeveloperCocoapodsExample/Podfile.lock
@@ -1,5 +1,5 @@
 PODS:
-  - Appcues (1.1.1)
+  - Appcues (1.1.2)
 
 DEPENDENCIES:
   - Appcues (from `../../Appcues.podspec`)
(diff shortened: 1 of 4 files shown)


Examples/DeveloperCocoapodsExample/Podfile.lock
```diff
@@ -9,8 +9,8 @@ EXTERNAL SOURCES:
     :path: "../../Appcues.podspec"
 
 SPEC CHECKSUMS:
-  Appcues: 758c5d86c18e49d09f501e771ca0cfd67651a5ce
+  Appcues: 6a4a6f7b5ebc2d0920b1a42421b732a185fc696b
 
 PODFILE CHECKSUM: 7a3abcd818687cb341a5696d302103f523e3edc0
 
-COCOAPODS: 1.11.3
+COCOAPODS: 1.11.2
```
iujames, line 16: looks like this is `1.11.3` in `main`, but still on later version in Gemfile in the `traits-2.0` branch. We can sort that out and update once things are rebased.

mmaatttt (approved):

PR #325, opened 2023-01-19
Use a weighted matching approach for element selector

iujames:
Still pondering best approach here, but this is an idea that uses a concept of weighted match - where some selector property matches are considered more valuable than others, and the largest weighted match value is considered the selected element if found. We could evolve what set of properties are used and how they are weighted over time.

This allows for multiple properties to be combined for a more distinct selector. If there is still no single highest value match, it fails element selection as before.


mmaatttt (approved):
Not sure that this will be the final approach, but at the very least, it's a good incremental improvement.

iujames:
> Not sure that this will be the final approach

Yeah, I think safe to say it will not be 😄

PR #327, opened 2023-01-30
Implement @appcues/tooltip trait UI

mmaatttt:
This implementation isn't entirely bulletproof yet, but it covers the basic use cases. It'll be easier in the future to have small PRs targeted at fixing specific scenarios.

A couple high-level things to note:

1. I updated the default and carousel containers to propagate the `preferredContentSize` so that the tooltip container can use it (the modal container uses constraints and so was able to size properly without this value).
2. In an unrelated change, I updated the target element and target rectangle traits to own `contentPreferredPosition` and `contentDistanceFromTarget`.
3. The value in using a generic `ExperienceWrapperViewController` for dialogs and tooltips is diminishing as the specific tooltip requirements increase. It might make sense to separate them instead of subclassing.


Sources/AppcuesKit/Presentation/Traits/Appcues/AppcuesTooltipTrait.swift
```diff
@@ -38,15 +45,20 @@ internal class AppcuesTooltipTrait: StepDecoratingTrait, WrapperCreatingTrait, P
     func createWrapper(around containerController: ExperienceContainerViewController) throws -> UIViewController {
         let dialogWrapperViewController = ExperienceWrapperViewController<TooltipWrapperView>(wrapping: containerController)
         dialogWrapperViewController.configureStyle(tooltipStyle)
+        dialogWrapperViewController.bodyView.pointerSize = hidePointer ? nil : pointerSize
```
iujames, line 48: Was `dialogWrapperViewController` maybe copied from something modal related? Wondering if there is a clearer name for this thing.

Sources/AppcuesKit/Presentation/Traits/Appcues/AppcuesTooltipTrait.swift
```diff
@@ -38,15 +45,20 @@ internal class AppcuesTooltipTrait: StepDecoratingTrait, WrapperCreatingTrait, P
...
+        if let preferredWidth = tooltipStyle?.width {
+            dialogWrapperViewController.bodyView.preferredWidth = preferredWidth
+        }
+        // Ensure content isnt hidden by the tooltip pointer
+        dialogWrapperViewController.bodyView.pointerInsetHandler = { insets in
+            containerController.additionalSafeAreaInsets = insets
+        }
```
iujames, line 55: just for my own learning, curious how this part works

Sources/AppcuesKit/Presentation/Traits/Appcues/ExperienceWrapperView.swift
```diff
@@ -11,61 +11,43 @@ import UIKit
...
-        }
-    }
+        contentWrapperView.translatesAutoresizingMaskIntoConstraints = false
 
-    func positionContentWrapperView() {
         NSLayoutConstraint.activate([
+            // this is required so the dialogView has an initial non-zero height, after which it can start sizing to the content.
+            contentWrapperView.layoutMarginsGuide.bottomAnchor.constraint(
+                greaterThanOrEqualTo: contentWrapperView.layoutMarginsGuide.topAnchor,
+                constant: 1),
```
iujames, line 49: came across this in something I was using earlier as well - is this specifically for the case of being embedded inside of the carousel and needing that initial height for the carousel sizing to work?

Sources/AppcuesKit/Presentation/Traits/Appcues/TooltipWrapperView.swift
```diff
@@ -11,38 +11,394 @@ import UIKit
...
+            targetFrame.origin.y = targetRectangle.maxY + distance
+            actualPosition = .bottom
         } else {
-            // default to a bottom anchored sheet
-            activeConstraints = [
-                // general position
-                contentWrapperView.bottomAnchor.constraint(equalTo: bottomAnchor),
-                contentWrapperView.leadingAnchor.constraint(equalTo: readableContentGuide.leadingAnchor),
-                contentWrapperView.trailingAnchor.constraint(equalTo: readableContentGuide.trailingAnchor),
+            // TODO: Should this consider a switch to horizontal if there's not enough vertical space on either side?
```
iujames, line 159: good question - does the current "shrink height if too tall to fit" mean it would result in scrolling content inside?

Sources/AppcuesKit/Presentation/Traits/Appcues/TooltipWrapperView.swift
```diff
@@ -11,38 +11,394 @@ import UIKit
...
+        // Leading triangle
+        if case .trailing = tooltipPosition {
+            tooltipPath.addLine(to: CGPoint(x: triangleRect.maxX, y: triangleRect.maxY))
+            tooltipPath.addLine(to: CGPoint(x: triangleRect.minX, y: triangleRect.midY))
+            tooltipPath.addLine(to: CGPoint(x: triangleRect.maxX, y: triangleRect.minY))
+        }
+
+        tooltipPath.close()
+
+        return tooltipPath.cgPath
```
iujames, line 387: holy smokes

iujames (approved):
A lot of heavy lifting there in TooltipWrapperView, and I can't pretend to fully follow all of that. But that is expected, and that complexity must go somewhere, and It seems well organized.

> The value in using a generic ExperienceWrapperViewController for dialogs and tooltips is diminishing

Agree, I don't think thats a big or meaningful loss to not have the shared code path there, unless you have any other considerations I'm missing.

Sources/AppcuesKit/Presentation/Traits/Appcues/AppcuesTooltipTrait.swift
mmaatttt: Good point, `dialog` is wrong here.

Sources/AppcuesKit/Presentation/Traits/Appcues/TooltipWrapperView.swift
mmaatttt: Yes, it would scroll

Sources/AppcuesKit/Presentation/Traits/Appcues/TooltipWrapperView.swift
mmaatttt: Just wait for the modifications for pointer edge cases 😅

Sources/AppcuesKit/Presentation/Traits/Appcues/TooltipWrapperView.swift
iujames: gotcha. I think it would be pretty cool if it could switch to leading or trailing tip in that case, but also think that seems totally fine to push down the road and see what feedback we can gather on it.

Sources/AppcuesKit/Presentation/Traits/Appcues/AppcuesTooltipTrait.swift
mmaatttt: The idea is that the space inside the pointer is part of the modal. This enabled a background image to fill the pointer space, and any scrolling content to scroll into the pointer space. 

L139 adds the height of the pointer to the view sizing. But, for example, if we have a top pointer, we don't want the content to start in the pointer space, so the right way to address that is with the safe area insets.

That leaves us in a place where the tooltip view needs to be able to modify the `additionalSafeAreaInsets` for the `containerController` based on where the pointer is determined to be positioned.  The challenge is the tooltip view doesn't know about the `containerController`. This closure bridges the gap by allowing the trait that creates the `ExperienceWrapperViewController` to provide the tooltip view a way how to modify the necessary insets (`didSet` on L30) when the new pointer position is determined (`setPointerInset()` on L275).

It's definitely convoluted, but it's the best approach I could think of.

Sources/AppcuesKit/Presentation/Traits/Appcues/ExperienceWrapperView.swift
mmaatttt: It's not carousel specific, but if the height is 0, SwiftUI seems to shortcut it layout process and it never gains the proper size.

Sources/AppcuesKit/Presentation/Traits/Appcues/AppcuesTooltipTrait.swift
iujames: makes sense! thanks for the details. It does feel like a proper approach to share that information in a necessarily decoupled way.

PR #328, opened 2023-01-30
Fix keyboard avoidance with the send capture dialog

iujames:
Three commits here to address the pieces

1. Extract out a protocol for the MultiLineTextView component to use for its model properties so that it could be shared with this use case, and we could get the "done" toolbar that is not available in SwiftUI in iOS 13
2. Update the SendCaptureUI dialog contents to use this text view implementation instead of the SwiftUI version
3. (the largest part) - rework the hosting ViewController structure to support loading our dialog contents, in SwiftUI, inside of a UIViewController that properly handles the keyboard avoidance and scrolling focused text view into visible region

The implementation of DebugModalViewController adapts ideas from the ExperienceWrapperViewController and ExperienceStepViewController, but condenses them down into a single ViewController structure, since the use case here is much more narrow. I think I've got this sorted out, but it was a little tricky.  Basic structure is this:

* DebugModalViewController has a root view of DebugModalWrapperView
* DebugModalWrapperView has two subviews - the backdropView which goes edge to edge, and the scrollView which is centered in the screen
* The scrollView has a contentView - which is where the embedded SwiftUI content ends up getting hosted
* The scrollView height is controlled by watching the preferredContentSize of that child SwiftUI hosted content (inside an AppcuesHostingViewController) 
* The DebugModalViewController can then listen for the necessary keyboard notifications and update contentInset on the scrollView as needed, in `adjustForKeyboard`

https://user-images.githubusercontent.com/19266448/215563283-74aab102-33af-4187-b1b4-b4ff74f28e9c.mov

### DIFF
--- a/Sources/AppcuesKit/Presentation/Debugger/ScreenCapturing/DebugModalViewController.swift
+++ b/Sources/AppcuesKit/Presentation/Debugger/ScreenCapturing/DebugModalViewController.swift
@@ -105,7 +105,7 @@ extension DebugModalViewController {
         let backdropView: UIView = {
             let view = UIView(frame: .zero)
             view.translatesAutoresizingMaskIntoConstraints = false
-            view.backgroundColor = UIColor(white: 0.0, alpha: 0.33)
+            view.backgroundColor = UIColor.label.withAlphaComponent(0.33)
             return view
         }()
 
@@ -116,6 +116,8 @@ extension DebugModalViewController {
             view.contentInsetAdjustmentBehavior = .always
             // For text input blocks, we want scrolling the modal content to be able to dismiss the keyboard.
             view.keyboardDismissMode = .interactive
+            view.layer.cornerRadius = 6.0
+            view.backgroundColor = UIColor.systemBackground
             return view
         }()
 
--- a/Sources/AppcuesKit/Presentation/Debugger/ScreenCapturing/DebugModalViewController.swift
+++ b/Sources/AppcuesKit/Presentation/Debugger/ScreenCapturing/DebugModalViewController.swift
@@ -1,0 +1,157 @@
+//
+//  DebugModalViewController.swift
+//  AppcuesKit
+//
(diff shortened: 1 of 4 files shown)


iujames:
@mmaatttt the latest commit here updates the handling of the items discussed previously - corner radius and scroll bar on scroll. Also reviewed and fixed dark mode handling.


https://user-images.githubusercontent.com/19266448/216114867-940b541b-e161-4cfb-8e29-862f95abf3da.mov

Sources/AppcuesKit/Presentation/Debugger/ScreenCapturing/DebugModalViewController.swift
```diff
@@ -0,0 +1,159 @@
...
+
+// A UIViewController that can be used to host any SwiftUI modal content to be presented,
+// and handle keyboard avoidance, if necessary.
+//
+// This is similar to the structure of the ExperienceWrapperViewController - providing
+// the backdrop - and the ExperienceStepViewController centered within - providing the scroll
+// container where the hosted SwiftUI content resides. However, this is a more specific and
+// reduced implementation that is decoupled from Experience step content and data.
+@available(iOS 13.0, *)
+internal class DebugModalViewController<Content>: UIViewController where Content: View {
```
mmaatttt, line 20: Is this different from `internal class DebugModalViewController<Content: View>: UIViewController {`?

I'm more used to seeing a generic type constraint after the parameter name.

mmaatttt (approved):
You're right that it might've been simpler entirely in UIKit 😅

Sources/AppcuesKit/Presentation/Debugger/ScreenCapturing/DebugModalViewController.swift
iujames: I believe totally equivalent, will update to match `ExperienceStepRootView` to keep consistent, thanks 👍

PR #330, opened 2023-01-31
Add dynamically adjusting tooltip pointer shapes

mmaatttt:
Supports 3 scenarios:

1. Regular, centred pointer
2. Right-angle pointer with a rounded corner
3. Right-angle pointer replacing a rounded corner

The `Triangle.init` is super long and complex, but it's really just the same logic 4x (one for each edge) with different values. Anything to make it more concise would incur a huge readability hit.

https://user-images.githubusercontent.com/845681/215852987-59a69af0-3100-4363-bf07-568fb0d6450c.mp4


iujames (approved):
Looks great! definitely one where a picture/video helps clarify a lot of code 😅

PR #331, opened 2023-02-02
Update tooltip layout to respect safe areas and always render fully on screen

mmaatttt:
Updating the tooltip positioning to handle safe areas and targets outside the screen bounds.

Also cleaned things up to always use `safeBounds` instead of referencing `layoutMargins` some places. This makes it easier to reason about what's happening.

Note that we might choose to throw errors from the `@appcues/target-rectangle` or `@appcues/target-element` trait if the target is entirely off the screen, but the tooltip trait should be able to render its content in a visible way regardless.

I will at some point create snapshot tests for every possible branch of the positioning code.

|Fixed|Old|
|-|-|
|fixed safe area top|bad safe area top|
|fixed leading|bad leading|
|good margin|bad margin|


iujames (approved):
nice

PR #334, opened 2023-02-07
Update sticky content approach

mmaatttt:
Adds a custom decoder for `Experience.Step.Child` which splits the `content` from the JSON into 3 different objects: `content`, `stickyTopContent`, and `stickyBottomContent`. Multiple sticky items get grouped into a vertical stack.

The `@appcues/sticky-content` trait is gone and the code from it is basically copied verbatim into `ExperienceStepViewController` (except I removed the `leading` and `trailing` options which never will be used).


Sources/AppcuesKit/Data/Models/Experience/ExperienceComponent+Sticky.swift
```diff
@@ -0,0 +1,102 @@
...
+// swiftlint:disable large_tuple
+
+extension ExperienceComponent {
+    /// Extract sticky stacks while filtering those from the body component.
+    func divided() -> (body: ExperienceComponent, stickyTop: ExperienceComponent?, stickyBottom: ExperienceComponent?) {
+        let (body, stickyTopComponents, stickyBottomComponents) = self.split()
+
+        return (
+            // `body` is `nil` only if `self` is sticky, which means there's no root content to be displayed so populate something empty.
+            body ?? ExperienceComponent.spacer(SpacerModel(id: UUID(), spacing: nil, style: nil)),
```
iujames, line 20: makes sense. I took a slightly different approach on Android where I split the content using the children of the root component (if it was a type that had children, which it is typically vstack) - effectively meaning there was always a root component that was assumed non-sticky. This is a bit more flexible, but I think probably both approaches work ok in practice.

Sources/AppcuesKit/Data/Models/Experience/ExperienceComponent+Sticky.swift
```diff
@@ -0,0 +1,102 @@
...
+}
+
+private extension Array where Element == ExperienceComponent {
+    /// Reduce multiple components into a single vertical stack.
+    func reduced() -> ExperienceComponent? {
+        switch count {
+        case 0:
+            return nil
+        case 1:
+            return self[0]
```
iujames, line 90: On Android, I always wrapped the sticky top/bottom regions in a vstack, if not empty, even in the single item case. This would be a bit more efficient in terms of overall depth of layout tree. I think we can update Android to match.

Sources/AppcuesKit/Presentation/UI/ExperienceStepViewController.swift
```diff
@@ -102,6 +110,42 @@ internal class ExperienceStepViewController: UIViewController {
             stepView.scrollView.scrollRectToVisible(frameInScrollView, animated: false)
         }
     }
+
+    private func decorateStickyContent(edge: ExperienceComponent.StackModel.StickyEdge, _ component: ExperienceComponent) {
+        // Must have the environmentObject so any actions in the sticky content can be applied.
```
iujames, line 115: just double checking - this gives it access to the form state as well, if there are form controls in sticky content, correct?

Sources/AppcuesKit/Presentation/UI/ExperienceStepViewController.swift
```diff
@@ -102,6 +110,42 @@ internal class ExperienceStepViewController: UIViewController {
...
+            constraints.append(stickyContentVC.view.bottomAnchor.constraint(equalTo: view.safeAreaLayoutGuide.topAnchor))
+        case .bottom:
+            constraints.append(stickyContentVC.view.topAnchor.constraint(equalTo: view.safeAreaLayoutGuide.bottomAnchor))
+        }
+
+        NSLayoutConstraint.activate(constraints)
+
+        stickyContentVC.didMove(toParent: self)
+
+        // Pass sticky content size changes to the parent controller to update the insets.
```
iujames, line 139: are these insets also how you find the boundary for the mask for fading content behind the sticky area in that poc?

iujames (approved):
nice clean implementation handling in the custom decoder! RIP sticky trait

Sources/AppcuesKit/Presentation/UI/ExperienceStepViewController.swift
mmaatttt: Yeah, exactly. That defines a mask with a frame sized from the safe area

Sources/AppcuesKit/Presentation/UI/ExperienceStepViewController.swift
mmaatttt: No actually! Thank you for checking, it's not working for a couple reasons:
1. Need the `stepState` environment object here like you noted
2. We only parse form items from the body content and the sticky items are removed from there before we parse so they're ignored entirely

Sources/AppcuesKit/Data/Models/Experience/ExperienceComponent+Sticky.swift
mmaatttt: Yeah, given our actual experiences having a root VStack, this doesn't matter. I could rewrite this as a `guard` to make it clearer that we really expect the `body` to be there and the Spacer is a last resort fallback. Clearer semantic than the nil-coalescing.

PR #335, opened 2023-02-09
Update animated GIF buffer calculation to allow larger caches

mmaatttt:
`maxOptimalSize` is the frame count that we buffer for an animated gif. Previously we'd compute the size of _all the frames in an image_ (`animatedImageDataSize`) to determine the cache size. That makes sense for the `if` condition where we check if the all the frames in the image should be cached.

That makes no sense for the subsequent `else if` check because the total number of frames in the gif is unrelated to the size of a single frame. I've updated it to determine how many frames we can cache while staying under the max cache size.

I've also updated that max size from 10MB to 15MB.

The result is this (a cache size of >1  should result in stutter-free playback, but a larger cache is safer)

|Image Resolution|Frame Count|Old Cached Frame Count|New Cached Frame Count|
|-|-|-|-|
|1080x1080|73|1|3|
|1600x1600|10|1|2|
|1600x1600|100|1|2|
|1620x1620|10|1|1|
|300x300|150|5|5|
|300x300|250|1|5|

You can see that `New Cached Frame Count` doesn't care about the `Frame Count`, and that the largest GIF that will play back smoothly is ~1600px square (we can increase this if we want, but that seems plenty large).

### DIFF
--- a/Sources/AppcuesKit/Vendor/FLAnimatedImage.swift
+++ b/Sources/AppcuesKit/Vendor/FLAnimatedImage.swift
@@ -255,15 +255,16 @@ internal class FLAnimatedImage: NSObject {
         if optimalFrameCacheSize == .noLimit {
             // Calculate the optimal frame cache size: try choosing a larger buffer window depending on the predicted image size.
             // It's only dependent on the image size & number of frames and never changes.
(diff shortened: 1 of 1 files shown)


iujames (approved):
nice update! non pressing, but can we contribute this back to FLAnimatedImage as a proposed change in any way? or raise an issue? Seems it would be a nice share for the community there.

mmaatttt:
I could open a PR on FLAnimatedImage, but there's been one release in the past 2 years, and what it's currently doing isn't wrong, it's just very conservative.

PR #336, opened 2023-02-09
Fix crash from lack of thread safety in SdkMetrics

mmaatttt:
The test fails reliably before the fix and never after the fix with hundreds of runs.

Note that `trackRender` is the only case of `syncQueue.sync`, the rest can be `async`.


iujames (approved):
nice find!

PR #337, opened 2023-02-13
Update sticky content to work using safe area

mmaatttt:
The previous approach was to inset the safe area of ExperienceStepViewController by the size of the sticky content and then set the sticky content outside the safe area. 

This was causing extra layout passes because SwiftUI knew the content was in the safe area and basically didn't appreciate that fact. It was slow in some cases and actually froze the app in one particularly extreme example that Ben found.

We also had an issue where the sticky content was taller than than available container height (especially notable in half sheets) and so the safe areas exceeded the container height and left no space for the content. This is user error in the experience design to a certain extent, but the way it was displayed on iOS looked even more broken. Now the sticky content will still entirely overlap, but in a way that makes sense.

Modifying just the scrollView insets allows the sticky content to be safely within the safe area. The only tricky part is needing to set the scrollview `contentOffset` as well to make sure the top of the content doesn't start under a sticky top element.

This branch was published as TestFlight build #55.

### DIFF
--- a/Sources/AppcuesKit/Presentation/UI/ExperienceStepViewController.swift
+++ b/Sources/AppcuesKit/Presentation/UI/ExperienceStepViewController.swift
@@ -24,6 +24,15 @@ internal class ExperienceStepViewController: UIViewController {
 
     private let contentViewController: UIViewController
 
+    var stickySpacing: UIEdgeInsets = .zero {
+        didSet {
+            stepView.scrollView.contentInset = stickySpacing
+            stepView.scrollView.scrollIndicatorInsets = stickySpacing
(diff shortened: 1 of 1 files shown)


iujames (approved):
good find and update - it's interesting to me, as it seems using `additionalSafeAreaInsets` and putting content above or below the safe area should be _valid_. Is this something unique to SwiftUI that does not support it?

mmaatttt:
It's not that it's _invalid_, it just has some awkward side effects. In UIKit you can write constraints to use (or not use) there safe area guides, but SwiftUI does it by default and doesn't seem to work really nicely otherwise. All that said, the main reason for changing was the container height issue.

PR #338, opened 2023-02-14
Add support for Authorization header on API requests

iujames:
This approach allows for passing an optional user signature during `identify` as a specific property name. 
```
Appcues.shared.identify(userID: userID, properties: ["appcues:user_id_signature": base64token])
```

The signature is a base64 encoded value that the customer has generated using their SDK key, outside of our SDK, in the customer app. It is then able to be used by our backend system to enforce that only authorized users can access content.

This is targeting `main` as it may need to release in a 1.3.x or 1.4 update before some of the longer running work.

[sc-48026]


Sources/AppcuesKit/Data/Networking/Authentication.swift
```diff
@@ -0,0 +1,28 @@
...
+//  AppcuesKit
+//
+//  Created by James Ellis on 2/14/23.
+//  Copyright © 2023 Appcues. All rights reserved.
+//
+
+import Foundation
+
+internal enum Authorization {
+    case bearer(String)
```
iujames, line 12: Some speculation here to encapsulate possibly other future auth types inside this type. In reality, we could probably flatten and remove, don't feel strongly.

mmaatttt (approved):
Change and tests look good!

Should we update any of the inline docs to mention the signature?

iujames:
> Should we update any of the inline docs to mention the signature?

Good idea I think - maybe a note on the `identify` API docs? Perhaps something in this doc as well https://github.com/appcues/appcues-ios-sdk/blob/main/Sources/AppcuesKit/AppcuesKit.docc/Identifying.md?plain=1#L5. Any others come to mind?

mmaatttt:
Both those spots seem good!

iujames:
@mmaatttt maybe I'll update API doc here, but then cycle back to the other documentation and examples. Reason - I want to ensure we have consistent messaging around this topic, like are we calling it "secure mode" like other products? or some other way to speak about it throughout our product line?

mmaatttt:
Sure, definitely fine to merge this PR and update the docs later.

PR #341, opened 2023-02-16
Screen capture API integration part 1 - infrastructure

iujames:
I'm working through the changes to integrate the new screen capture APIs. This PR attempts to break out the first phase of what I'm calling "infrastructure improvements", so as to not have too gigantic of a combined PR.

The updates here
* Add support for new capabilities needed in NetworkClient
    * Support POST requests and responses with no body
    * Support PUT request (with empty response body)
    * Support optional Authorization header
* Add new response models that will be used in the upcoming network flow
* Add endpoints for calls that will be used in the upcoming network flow

And touch up the existing tests to continue to function.

Follow-on updates will be
* Updating the screen capture deep link handling to include the JWT
* Implementing the actual network flow to save a screenshot and screen, and hooking into debugger UI
* Adding additional tests for these updates


Sources/AppcuesKit/Data/Networking/NetworkClient.swift
```diff
@@ -91,11 +154,43 @@ internal class NetworkClient: Networking {
 
         dataTask.resume()
     }
+
+    // version that does not decode any response object, assumes empty or discards
+    private func handleRequest(_ urlRequest: URLRequest,
```
iujames, line 159: some redundancy here with `handleRequest<T>` but I couldn't think of a great way to combine them that didn't make it more complicated and harder to read 🤷‍♂️ Open to ideas if this is problematic.

Sources/AppcuesKit/Data/Networking/Endpoint.swift
```diff
@@ -42,6 +42,58 @@ internal enum APIEndpoint: Endpoint {
...
+            components.path = "/v1/accounts/\(config.accountID)/mobile/\(config.applicationID)/pre-upload-screenshot?name=\(filename)"
+        case .screenCapture:
+            components.path = "/v1/accounts/\(config.accountID)/mobile/\(config.applicationID)/screens"
+        }
+
+        return components.url
+    }
+}
+
+internal struct URLEndpoint: Endpoint {
```
iujames, line 89: this is the one that will be used for by the pre-signed image upload path - S3 path, will comment and have usage in next PR.

Sources/AppcuesKit/Data/Models/ScreenshotUpload.swift
```diff
@@ -0,0 +1,21 @@
...
+//  AppcuesKit
+//
+//  Created by James Ellis on 2/15/23.
+//  Copyright © 2023 Appcues. All rights reserved.
+//
+
+import Foundation
+
+// response model for the customer API pre-upload-screenshot endpoint
+internal struct ScreenshotUpload: Decodable {
```
mmaatttt, line 12: I think I like the idea of grouping the all screen capture stuff _somewhere_. Maybe even a top level directory alongside Data/Presentation/etc? Not sure the right home, but I don't like the idea of having it scattered throughout.

mmaatttt (approved):

Sources/AppcuesKit/Data/Models/ScreenshotUpload.swift
iujames: Yeah, good call. I'll look into structure clean up there.

iujames: moved them all into `/Presentation/Debugger/ScreenCapturing` for now, to be together at least. This is in the part 3 PR on top of this in commit https://github.com/appcues/appcues-ios-sdk/pull/343/commits/b851302eb4d4a820fd512a91e633df49288ded4f -- will squash these down.

PR #342, opened 2023-02-16
Screen capture API integration part 2 - auth token on screen capture deep link

iujames:
stacks on #341 

This one adds support for the `token` parameter on the screen capture deep link, and passes it through to the spot where we'll need it once we hook in the capture logic.

Deep link structure is now like
```
appcues-{app_id}://sdk/capture_screen?token=jwt
```


Sources/AppcuesKit/Presentation/Debugger/UIDebugger.swift
```diff
@@ -138,14 +138,14 @@ extension UIDebugger: DebugViewDelegate {
...
-            captureScreen()
+        case let .screenCapture(authorization):
+            captureScreen(authorization: authorization)
         case .show, .close, .reposition:
             break
         }
     }
 
-    private func captureScreen() {
+    private func captureScreen(authorization: Authorization) {
```
iujames, line 148: this function is where things will hook in later and use the auth to make the customer API calls

Sources/AppcuesKit/Presentation/DeepLinkHandler.swift
```diff
@@ -46,8 +46,8 @@ internal class DeepLinkHandler: DeepLinkHandling {
                 self = .debugger(destination: DebugDestination(pathToken: pathTokens[safe: 1]))
             } else if pathTokens.count == 2, pathTokens[0] == "verify" {
                 self = .verifyInstall(id: pathTokens[1])
```
mmaatttt, line 48: Side note, a query param seems "more right" for the verify deeplink check too. I might update that at some point.

Sources/AppcuesKit/Presentation/Debugger/FloatingView/DebugView.swift
```diff
@@ -495,6 +495,6 @@ extension DebugView {
         case open
         case close
         case reposition
-        case screenCapture
+        case screenCapture(Authorization)
```
mmaatttt, line 498: I suspect we're going to want a new `JWT` struct instead of `Authorization` here eventually since we're going to need to decode the JWT to get the expiry date. Obviously this is fine until the need arises.

mmaatttt (approved):

Sources/AppcuesKit/Presentation/Debugger/FloatingView/DebugView.swift
iujames: Perhaps, I was hoping to avoid that honestly - do we need to know? What would we do with that? Maybe we can just detect the 403 and give an appropriate error message to say re-establish your session?

mmaatttt: The main thing I think we talked about (I think when you were away) was automatically hiding the capture button once the token is expired. A minor quality of life improvement versus capturing and getting an error, so definitely not required.

It's quite simple to decode just the value we need without a library or anything (something like [this](https://stackoverflow.com/a/40915703)) so it wouldn't be a huge lift to do.

iujames: I see, yeah that would be a nice to have. Thanks

PR #344, opened 2023-02-17
Feature/skippable trait enhancements

mmaatttt:
Adds new config options to `@appcues/skippable`. The minimal close button draws a thinner x using a `differenceBlendMode`. The button target is still the same size for usability.

I also updated a few trait inits to match the schemas: if no config values are required in the spec, then the trait should init properly without requiring a config dictionary. (Noticed this because I was doing skippable wrong)

### DIFF
--- a/Sources/AppcuesKit/Presentation/Traits/Appcues/AppcuesTooltipTrait.swift
+++ b/Sources/AppcuesKit/Presentation/Traits/Appcues/AppcuesTooltipTrait.swift
@@ -27,10 +27,10 @@ internal class AppcuesTooltipTrait: StepDecoratingTrait, WrapperCreatingTrait, P
     let pointerSize: CGSize
 
     required init?(configuration: ExperiencePluginConfiguration, level: ExperienceTraitLevel) {
-        guard let config = configuration.decode(Config.self) else { return nil }
-        self.hidePointer = config.hidePointer ?? false
-        self.pointerSize = CGSize(width: config.pointerBase ?? 16, height: config.pointerLength ?? 8)
-        self.tooltipStyle = config.style
+        let config = configuration.decode(Config.self)
+        self.hidePointer = config?.hidePointer ?? false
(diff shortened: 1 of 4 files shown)


Sources/AppcuesKit/Presentation/Traits/Appcues/AppcuesTooltipTrait.swift
```diff
@@ -27,10 +27,10 @@ internal class AppcuesTooltipTrait: StepDecoratingTrait, WrapperCreatingTrait, P
     let pointerSize: CGSize
 
     required init?(configuration: ExperiencePluginConfiguration, level: ExperienceTraitLevel) {
-        guard let config = configuration.decode(Config.self) else { return nil }
-        self.hidePointer = config.hidePointer ?? false
-        self.pointerSize = CGSize(width: config.pointerBase ?? 16, height: config.pointerLength ?? 8)
-        self.tooltipStyle = config.style
+        let config = configuration.decode(Config.self)
+        self.hidePointer = config?.hidePointer ?? false
```
iujames, line 31: should we note this default `false` [here](https://github.com/appcues/appcues-mobile-experience-spec/blob/main/trait-schemas/appcues-tooltip-schema.json#L21)? It does seem pretty obvious in this case that it would not be hidden by default for a standard tooltip.

iujames (approved):
looking good! nice to get all sync'ed up with spec

Sources/AppcuesKit/Presentation/Traits/Appcues/AppcuesTooltipTrait.swift
mmaatttt: Yeah, explicitness is always good. We have an unwritten rule that our config booleans are named in such a way that they're option and default to false, but good to be clear about it!

mmaatttt: https://github.com/appcues/appcues-mobile-experience-spec/pull/140

PR #345, opened 2023-02-17
Do not mark experience completed if dismissed on the final step

iujames:
Requested change for parity with web, and to make our Studio analytics dashboards more comprehendible for customers.

### DIFF
--- a/Tests/AppcuesKitTests/Experiences/ExperienceRendererTests.swift
+++ b/Tests/AppcuesKitTests/Experiences/ExperienceRendererTests.swift
@@ -14,15 +14,10 @@ class ExperienceRendererTests: XCTestCase {
 
     var appcues: MockAppcues!
     var experienceRenderer: ExperienceRenderer!
-    private var updates: [TrackingUpdate] = []
 
     override func setUpWithError() throws {
         appcues = MockAppcues()
         experienceRenderer = ExperienceRenderer(container: appcues.container)
-    }
-
-    override func tearDownWithError() throws {
-        updates = []
     }
 
     func testShowPublished() throws {
@@ -260,13 +255,14 @@ class ExperienceRendererTests: XCTestCase {
         // Arrange
         let completionExpectation = expectation(description: "Completion called")
 
+        var updates: [TrackingUpdate] = []
         let preconditionPresentExpectation = expectation(description: "Experience presented")
         let dismissExpectation = expectation(description: "Experience dismissed")
         let experience = ExperienceData.singleStepMock
         let preconditionPackage: ExperiencePackage = experience.package(presentExpectation: preconditionPresentExpectation, dismissExpectation: dismissExpectation)
         appcues.traitComposer.onPackage = { _, _ in preconditionPackage }
         experienceRenderer.show(experience: ExperienceData(experience.model, trigger: .showCall, priority: .low, published: true), completion: nil)
-        appcues.analyticsPublisher.onPublish = { update in self.updates.append(update) }
+        appcues.analyticsPublisher.onPublish = { update in updates.append(update) }
         wait(for: [preconditionPresentExpectation], timeout: 1)
 
         // Act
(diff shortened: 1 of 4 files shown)


Tests/AppcuesKitTests/Experiences/ExperienceRendererTests.swift
```diff
@@ -251,6 +256,31 @@ class ExperienceRendererTests: XCTestCase {
...
+        // Act
+        experienceRenderer.dismissCurrentExperience(markComplete: false) { _ in
+            completionExpectation.fulfill()
+        }
+
+        // Assert
+        waitForExpectations(timeout: 1)
+        let lastUpdate = try XCTUnwrap(updates.last)
+        // confirm that dismiss on last step triggers experience_dismissed analytics, not experience_completed
+        XCTAssertEqual(lastUpdate.type, .event(name: "appcues:v2:experience_dismissed", interactive: false))
```
iujames, line 281: Bit of a weird test here where I'm actually relying on analytics to confirm the behavior of the ExperienceRenderer logic, but I wasn't able to find a better way. We did not have a test in this area previously.

Tests/AppcuesKitTests/Experiences/ExperienceStateMachineTests.swift
```diff
@@ -176,6 +181,34 @@ class ExperienceStateMachineTests: XCTestCase {
...
+        // Act
+        (package.containerController as! Mocks.ContainerViewController).mockIsBeingDismissed = true
+        package.containerController.viewWillDisappear(false)
+        package.containerController.viewDidDisappear(false)
+
+        // Assert
+        waitForExpectations(timeout: 1)
+        let lastUpdate = try XCTUnwrap(updates.last)
+        // confirm that dismiss on last step triggers experience_dismissed analytics, not experience_completed
+        XCTAssertEqual(lastUpdate.type, .event(name: "appcues:v2:experience_dismissed", interactive: false))
```
iujames, line 209: And another test where we're relying on the AnalyticsObserver to confirm the logic behavior in the StateMachine, but worked nicely for this case.

Tests/AppcuesKitTests/Experiences/ExperienceStateMachineTests.swift
```diff
@@ -176,6 +181,34 @@ class ExperienceStateMachineTests: XCTestCase {
...
 
+    func test_stateIsRenderingStep_onViewControllerDismissed_doesNotMarkComplete() throws {
+        // the @appcues/skippable trait would do this
+
+        // Arrange
+        let dismissExpectation = expectation(description: "Experience dismissed")
+        let experience = ExperienceData.singleStepMock
+        let package: ExperiencePackage = experience.package(dismissExpectation: dismissExpectation)
+
+        let initialState: State = .renderingStep(experience, Experience.StepIndex(group: 0, item: 0), package, isFirst: true)
```
mmaatttt, line 192: Should this be tested from the last step in the Mock experience (`group: 0, item: 0`) to confirm that skipping the last step gets us the `experience_dismissed` analytic?

Tests/AppcuesKitTests/Experiences/ExperienceRendererTests.swift
```diff
@@ -14,12 +14,17 @@ class ExperienceRendererTests: XCTestCase {
 
     var appcues: MockAppcues!
     var experienceRenderer: ExperienceRenderer!
+    private var updates: [TrackingUpdate] = []
```
mmaatttt, line 17: If `updates` is only used in the new test case, it might be simplest to have it in that function only?

mmaatttt (approved):

Tests/AppcuesKitTests/Experiences/ExperienceStateMachineTests.swift
iujames: do you mean `mock` instead of `singleStepMock` and use (`group: 1, iem: 0`) ?

Tests/AppcuesKitTests/Experiences/ExperienceRendererTests.swift
iujames: yes, should have cleaned that up, will do.

PR #352, opened 2023-03-01
Implement screen capture error handling and retry

iujames:
This adds the pink failure toast and the button to retry the save of the screen.

Toast itself was pretty simple, but this required reworking how tap handling was done, since previously all taps were just intercepted by the host UIView and used to dismiss the toast.

Now, the SwiftUI side needs to be able to handle taps, for the retry button, and needs a way to support dismissing its host `UIHostingController`. This is where the new `ToastHostingController` and `ToastHostProvider` come into play. The provider is injected into the SwiftUI view through an Environment object, and gives access to a `dismissToast()` function that the SwiftUI side can invoke. This allows the caller to set up a closure to actually do whatever work is necessary to dismiss - in this case un-embedding the child VC and pulling this toast out of the view. The general idea here on how to get access to the parent controller from inside the SwiftUI View came from this post https://augmentedcode.io/2022/09/19/accessing-uihostingcontroller-from-a-swiftui-view/ .

Again, things got a bit more complicated here due to the fact that this is using SwiftUI inside of a hosted UIViewController for the debugger views.

Failure:

https://user-images.githubusercontent.com/19266448/222171732-aed7d432-4986-460b-8393-ce9f60911f6b.mov

Success: 

https://user-images.githubusercontent.com/19266448/222171776-b9ea58b8-5750-48c8-97d6-24353036b9d6.mov


Sources/AppcuesKit/Presentation/Debugger/ScreenCapturing/ToastHostingController.swift
```diff
@@ -0,0 +1,58 @@
...
+            onDismissToast?(animated)
+        }
+    }
+}
+
+// Helper that embeds the given toast View and provides access to the ToastHosting implementation
+// in a ToastHostProvider environment object.
+@available(iOS 13.0, *)
+internal extension View {
+    func embeddedInToastHostingController() -> ToastHostingController<some View> {
```
mmaatttt, line 50: Wow this has has taken me a bit to wrap my head around. Has me wondering if it'd be easier to do all the toast stuff in UIKit.

mmaatttt (approved):

Sources/AppcuesKit/Presentation/Debugger/ScreenCapturing/ToastHostingController.swift
iujames: Yeah I think it would be. And I'd be interested in taking another pass at it to try and see how much of this can be ripped out.

iujames: will close this PR if we agree to go forward with UIKit alternate approach instead, in https://github.com/appcues/appcues-ios-sdk/pull/357

iujames:
closing in favor of #357

PR #355, opened 2023-03-02
Apply step padding to sticky content

mmaatttt:
Having to manage the scroll insets separately now is a bit annoying, but I haven't found a nicer solution.

Example with 10px padding on all sides, no margins on anything, and the image/button sticky top/bottom respectively:

ezgif-4-b53ee13e93


iujames (approved):
I think this is good. 

I don't know why but for some reason I had the instinct that the scroll indicator should still go top to bottom, and just allow underscroll in the sticky area, like it was part of the safe area that can be scrolled behind and the main content is edge to edge? But in this case the content height is not the full height. I assume it is scrolling under, just obscured by the white background in the bottom button row in this case right?

PR #356, opened 2023-03-07
Add tests for tooltip frame calculations

mmaatttt:
And the tests caught a layout issue for the bottom sheet cases (when there's no target rectangle available).

### DIFF
--- a/Tests/AppcuesKitTests/Experiences/TooltipWrapperViewFrameTests.swift
+++ b/Tests/AppcuesKitTests/Experiences/TooltipWrapperViewFrameTests.swift
@@ -1,0 +1,399 @@
+//
+//  TooltipWrapperViewFrameTests.swift
+//  AppcuesKitTests
+//
+//  Created by Matt on 2023-03-03.
+//  Copyright © 2023 Appcues. All rights reserved.
+//
+
+import XCTest
+@testable import AppcuesKit
+
+class TooltipWrapperViewFrameTests: XCTestCase {
+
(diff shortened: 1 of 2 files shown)


Tests/AppcuesKitTests/Experiences/TooltipWrapperViewFrameTests.swift
```diff
@@ -0,0 +1,399 @@
...
+
+        // Act
+        view.positionContentView()
+
+        // Assert
+        XCTAssertEqual(
+            view.shadowWrappingView.frame,
+            CGRect(x: 8, y: 552, width: 377, height: 300),
+            "Inside the layout margins and at the bottom of the screen"
+        )
```
iujames, line 44: Note to self and @andretortolano, I think we may still need handling around the default frame on Android in the no target rectangle case, to match.

iujames (approved):

Tests/AppcuesKitTests/Experiences/TooltipWrapperViewFrameTests.swift
iujames: @mmaatttt refresh memory please, is this for the case where an `@appcues/tooltip` trait is used, but no `@appcues/target-rectangle` or `@appcues/target-element` exists at all (or any other rect metadata setting trait)?

In the case of a failed target-element trait, we do not fall back to a bottom anchored tooltip, correct? But rather a fatal experience error and end of the experience.

mmaatttt: Yes and yes! 👍

PR #357, opened 2023-03-08
Add success and failure toast implementations for screen capture using simplified UIView

iujames:
This is a second swing at what was done in #352 , using a simplified approach that implements the toasts in a UIView, rather than having to [deal with all the complexities](https://github.com/appcues/appcues-ios-sdk/pull/352#discussion_r1126649047) of hosting a SwiftUI toast view inside of a UIKit container.

The end result should look and behave identically to the previous PR.
* Toast shows 3 sec on success, 6 sec on failure
* Tapping retry button triggers another save attempt with the same capture and token (useful in rare client network glitch case)
* Tapping anywhere else outside of the toast, or any part of the toast other than the button will initiate a dismiss

Success

https://user-images.githubusercontent.com/19266448/223795054-e1f82f63-6455-4332-a1b5-ad4e3b247351.mov

Failure

https://user-images.githubusercontent.com/19266448/223795086-f91ea818-cb2f-47ab-af5b-caa44ae07add.mov

### DIFF
--- a/Sources/AppcuesKit/Presentation/Debugger/FloatingView/DebugView.swift
+++ b/Sources/AppcuesKit/Presentation/Debugger/FloatingView/DebugView.swift
@@ -23,6 +23,7 @@ internal class DebugView: UIView {
 
     private let floatingViewPanRecognizer = UIPanGestureRecognizer()
     private let backgroundTapRecognizer = UITapGestureRecognizer()
+    private let toastTapRecognizer = UITapGestureRecognizer()
 
     private var initialTouchOffsetFromCenter: CGPoint = .zero
 
@@ -74,8 +75,8 @@ internal class DebugView: UIView {
         return view
     }()
 
-    var toastWrapperView: UIView = {
-        let view = UIView()
+    var toastView: CaptureToastView = {
+        let view = CaptureToastView()
         view.translatesAutoresizingMaskIntoConstraints = false
         view.alpha = 0
 
@@ -140,7 +141,7 @@ internal class DebugView: UIView {
         addSubview(panelWrapperView)
         addSubview(dismissView)
         addSubview(floatingView)
-        addSubview(toastWrapperView)
+        addSubview(toastView)
 
         backgroundView.pin(to: self)
         NSLayoutConstraint.activate([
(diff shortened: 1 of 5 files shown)


Sources/AppcuesKit/Presentation/Debugger/FloatingView/DebugView.swift
```diff
@@ -153,17 +154,21 @@ internal class DebugView: UIView {
             dismissView.trailingAnchor.constraint(equalTo: trailingAnchor),
             dismissView.bottomAnchor.constraint(equalTo: bottomAnchor),
 
-            toastWrapperView.leadingAnchor.constraint(equalTo: leadingAnchor),
-            toastWrapperView.trailingAnchor.constraint(equalTo: trailingAnchor),
-            toastWrapperView.bottomAnchor.constraint(equalTo: bottomAnchor)
+            toastView.leadingAnchor.constraint(equalTo: leadingAnchor, constant: 25),
+            toastView.trailingAnchor.constraint(equalTo: trailingAnchor, constant: -25),
```
mmaatttt, line 158: Maybe use the `readableContentGuide` here so we don't get super wide toasts on iPads?

Sources/AppcuesKit/Presentation/Debugger/ScreenCapturing/CaptureToastView.swift
```diff
@@ -0,0 +1,103 @@
...
+        stackView.addArrangedSubview(messageLabel)
+        stackView.addArrangedSubview(retryButton)
+
+        addSubview(stackView)
+
+        NSLayoutConstraint.activate([
+            stackView.leadingAnchor.constraint(equalTo: leadingAnchor, constant: 16),
+            stackView.trailingAnchor.constraint(equalTo: trailingAnchor, constant: -16),
+            stackView.topAnchor.constraint(equalTo: topAnchor, constant: 12),
+            stackView.bottomAnchor.constraint(equalTo: bottomAnchor, constant: -12),
```
mmaatttt, line 55: This is fine, but in case you didn't know, you can set `layoutMargins` on the stack view and then set `stackView.isLayoutMarginsRelativeArrangement = true`. Which approach you take really is just dependent on how you visualize the view in your mind. One slight benefit of the layout margins is not having the extra cognitive load of "why is this constant value negative again?"

mmaatttt (approved):
Nice, I definitely prefer this approach! Thanks for iterating on it.

Sources/AppcuesKit/Presentation/Debugger/FloatingView/DebugView.swift
iujames: yes, thank you - had a pending comment about tablet and landscape that i forgot to submit, doh. but yeah exactly, will do.

Sources/AppcuesKit/Presentation/Debugger/ScreenCapturing/CaptureToastView.swift
iujames: nice, yeah that is probably a cleaner route here, will try it out, thanks.

Sources/AppcuesKit/Presentation/Debugger/FloatingView/DebugView.swift
iujames: OK I think this looks reasonable now:

https://user-images.githubusercontent.com/19266448/224070636-699ef9f7-e44d-4515-a4d6-43f4298e8148.mov

PR #360, opened 2023-03-09
Fix keyboard avoidance for small bottom-aligned modals

mmaatttt:
This approach preserves the "scroll to visible" approach if the bottom aligned modal is tall enough, otherwise it shifts the whole modal up to ensure the content is visible.

Animation for the shift on the short modal isn't included because the shadow doesn't properly animate. This will be fixed in the future.

## Tall
https://user-images.githubusercontent.com/845681/224135151-c8ec12f7-17eb-45c3-b35f-0dd48df8861a.mp4

## Short 
https://user-images.githubusercontent.com/845681/224135170-24141b21-d45c-4db5-8ed1-e6a08225fa0d.mp4

### DIFF
--- a/Sources/AppcuesKit/Presentation/Traits/Appcues/DialogContainerViewController.swift
+++ b/Sources/AppcuesKit/Presentation/Traits/Appcues/DialogContainerViewController.swift
@@ -23,6 +23,9 @@ internal class DialogContainerViewController: UIViewController {
         modalPresentationStyle = .overFullScreen
     }
 
+    // Only set for bottom-aligned modals. Need access to the constant value for keyboard avoidance.
+    var bottomConstraint: NSLayoutConstraint?
+
     @available(*, unavailable)
     required init?(coder: NSCoder) {
         fatalError("init(coder:) has not been implemented")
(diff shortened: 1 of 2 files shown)


Sources/AppcuesKit/Presentation/Traits/Appcues/DialogContainerViewController.swift
```diff
@@ -53,11 +71,31 @@ internal class DialogContainerViewController: UIViewController {
...
+            self.bottomConstraint?.constant = 0
+        } else {
+            switch containerView.dialogView.keyboardAvoidanceStrategy(keyboardFrameInScreen: keyboardValue.cgRectValue) {
+            case .scroll:
+                // The ExperienceStepViewController handles this case and scrolls the content to avoid the keyboard.
+                break
+            case .move(let keyboardFrame):
+                // Not enough room to scroll the firstResponder into view, so adjust the position of the dialogView.
+                self.bottomConstraint?.constant = -keyboardFrame.height
+            }
```
iujames, line 98: curious, how does this still give padding above the keyboard and not just sit directly on top of the keyboard frame like in the previous video in slack?

iujames (approved):
looks good - only thought is to perhaps check something really short like the iphone SE, if you have not already. Perhaps landscape, just though about that potential complication.

Sources/AppcuesKit/Presentation/Traits/Appcues/DialogContainerViewController.swift
mmaatttt: I'm using `keyboardFrame.height` here instead of `intersection.height`. The intersection is smaller because the safe area at the bottom is excluded. I was using `intersection` because that's what I copied from the scrolling implementation (which correctly uses `intersection`), but that's wrong in this case.

mmaatttt: And regarding landscape: it's not pretty (landscape rarely is), but it works as expected.

PR #366, opened 2023-03-16
Add support for rounded tooltip pointers

mmaatttt:
https://stackoverflow.com/q/20442203 is an exceptionally helpful post.

I used the approach from [this answer](https://stackoverflow.com/a/20646446) to actually implement the rounding (hence the need to change from a UIBezierPath to CGMutablePath for access to `addArc(tangent1End:tangent2End:radius:)`).

The visualizations and code from [this answer](https://stackoverflow.com/a/20644065) were crucial for helping me figure out the max corner radius calculation. I've tried to extensively document what `maxPointerCornerRadius` is doing since it's not the slightest bit easy to understand. I've attached a playground with a visualization that might be insightful.

Screenshot 2023-03-16 at 4 18 49 PM

The second commit here renames the now ambiguous `cornerRadius`, so if you want to skip the noise, review the first commit by itself.

[Pointer Radius Visualization.playground.zip](https://github.com/appcues/appcues-ios-sdk/files/10995550/Pointer.Radius.Visualization.playground.zip)

### DIFF
--- a/Sources/AppcuesKit/Presentation/Extensions/UIBezierPath+Tooltip.swift
+++ b/Sources/AppcuesKit/Presentation/Extensions/UIBezierPath+Tooltip.swift
@@ -17,17 +17,20 @@ internal struct Pointer {
     let size: CGSize
     let cornerRadius: CGFloat
     let offset: CGFloat
+
+    func pathInfo(mainRect: CGRect, boxCornerRadius: CGFloat) -> Pointer.PathInfo {
+        PathInfo(pointer: self, mainRect: mainRect, boxCornerRadius: boxCornerRadius)
+    }
 }
 
 extension CGPath {
     static func tooltip(around mainRect: CGRect, boxCornerRadius: CGFloat, pointer: Pointer) -> CGPath {
         let path = CGMutablePath()
-
-        let triangle = CGMutablePath.Triangle(pointer: pointer, mainRect: mainRect, boxCornerRadius: boxCornerRadius)
+        let pointerPathInfo = pointer.pathInfo(mainRect: mainRect, boxCornerRadius: boxCornerRadius)
 
         // Draw the path clockwise from top left
 
-        if !triangle.overridesTopLeftCorner {
+        if !pointerPathInfo.overridesTopLeftCorner {
             let topLeft = CGPoint(x: mainRect.minX + boxCornerRadius, y: mainRect.minY + boxCornerRadius)
             path.addRelativeArc(center: topLeft, radius: boxCornerRadius, startAngle: .pi, delta: .pi / 2)
         } else {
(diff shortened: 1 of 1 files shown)


Sources/AppcuesKit/Presentation/Extensions/UIBezierPath+Tooltip.swift
```diff
@@ -15,84 +15,118 @@ internal struct Pointer {
...
-    convenience init(tooltipAround mainRect: CGRect, cornerRadius: CGFloat, pointer: Pointer) {
-        self.init()
+extension CGPath {
+    static func tooltip(around mainRect: CGRect, cornerRadius: CGFloat, pointer: Pointer) -> CGPath {
+        let path = CGMutablePath()
 
-        let triangle = Triangle(pointer: pointer, mainRect: mainRect, cornerRadius: cornerRadius)
+        let triangle = CGMutablePath.Triangle(pointer: pointer, mainRect: mainRect, cornerRadius: cornerRadius)
 
         // Draw the path clockwise from top left
```
iujames, line 28: I noticed all the `clockwise` params switched from true to false, is this still an accurate comment?

Sources/AppcuesKit/Presentation/Extensions/UIBezierPath+Tooltip.swift
```diff
@@ -15,84 +15,118 @@ internal struct Pointer {
...
+        let point0: CGPoint
+        let point4: CGPoint
+
         private(set) var overridesTopLeftCorner: Bool
         private(set) var overridesTopRightCorner: Bool
         private(set) var overridesBottomRightCorner: Bool
         private(set) var overridesBottomLeftCorner: Bool
 
+        private(set) var offCenterPointer1: Bool
+        private(set) var offCenterPointer2: Bool
```
iujames, line 116: curious, what do these represent exactly?

iujames (approved):
wow! I'm impressed

Sources/AppcuesKit/Presentation/Extensions/UIBezierPath+Tooltip.swift
mmaatttt: We're still drawing clockwise in terms of the order of sides and corners. It seems like CGPath draws differently from UIBezierPath, and so just changing this value made it work, but I'm not sure I can explain why. I was going to revisit these angles but forgot. Will take a closer look.

mmaatttt: Yeah, UIKit and Quartz have different coordinate systems. UIKit left, Quartz right:
image

https://www.informit.com/articles/article.aspx?p=2149190&seqNum=11

I also noticed `addRelativeArc` is an option that's simpler to use, so switching to that since it doesn't have clockwise options at all. Still not sure why `clockwise: false` was correct.

Sources/AppcuesKit/Presentation/Extensions/UIBezierPath+Tooltip.swift
mmaatttt: Updated the naming to hopefully be clearer and also took the opportunity to rename the `Triangle` struct to be `Pointer.PathInfo` for clarity.

Sources/AppcuesKit/Presentation/Extensions/UIBezierPath+Tooltip.swift
iujames: ah ok, didn't know that. thanks for the info.

Sources/AppcuesKit/Presentation/Extensions/UIBezierPath+Tooltip.swift
iujames: thanks - updated names help I think. Adding that explanation above in the comments maybe helpful for future ref as well.

PR #370, opened 2023-03-22
Support configurable element targeting strategy and default UIKit implementation

iujames:
This adds the capability to supply an `AppcuesElementTargeting` implementation into the config at initialization, which supports extensibility for other strategies around capturing screens and targeting/selecting elements. By default, it uses a UIKit/UIView "standard" implementation. But, other uses like the React Native plugin, for example, can supply their own.

This adds a few types to the public API to support this, and the `@objc` compatibility support for those made some slightly different than initially anticipated - i.e. a class instead of a protocol, for example.

* `AppcuesElementTargeting` - root protocol for the "strategy"
* `AppcuesElementSelector` - class (which originally was a protocol before objc) to define selector matching behavior
* `AppcuesViewElement` - the layout metadata for a View, used in capture and target element rendering


Sources/AppcuesKit/Appcues.swift
```diff
@@ -12,6 +12,10 @@ import Foundation
 @objc(Appcues)
 public class Appcues: NSObject {
 
+    // internal static, made available to the AppcuesTargetElementTrait
+    @available(iOS 13.0, *)
+    static var elementTargeting: AppcuesElementTargeting = UIKitElementTargeting()
```
iujames, line 17: Open to ideas here. 

Other options considered were somehow passing this through from the `TraitComposer`, using  some mechanism to identify which traits needed access to it (i.e. a protocol or certain type); passing the `Appcues` instance through to traits somehow like we do Actions.

That last one seemed quite invasive in particular, for this special case need. But somehow, the `AppcuesTargetElementTrait` needs access during rendering, and the `UIDebugger` needs access during screen capture.

Sources/AppcuesKit/Presentation/Public/AppcuesElementTargeting.swift
```diff
@@ -0,0 +1,35 @@
...
+    ///            the given properties. If no applicable properties are found, the return value should be `nil`.
+    func inflateSelector(from properties: [String: String]) -> AppcuesElementSelector?
+}
+
+internal extension AppcuesElementTargeting {
+    // Default implementation that looks at the current view capture and finds any matches.
+    // Used during targeted element experiences to find the position in the UI for the experience.
+    func findMatches(for selector: AppcuesElementSelector) -> [AppcuesViewElement] {
+        return captureLayout()?.viewsMatchingSelector(selector) ?? []
+    }
```
iujames, line 34: Calling out that this is a bit less peformant than before. It now traverses the view tree once to capture the layout, then walks that tree a second time to find the views matching the selector. This could be combined into a single traversal of the view trying to find selector matches, but that would shift responsibility into the AppcuesElementTargeting implementation, rather than just using the layout capture it already provides in the internal spot here. I went this route to start, to lessen the implementation burden of any custom targeting strategy.

Sources/AppcuesKit/Presentation/Traits/Appcues/AppcuesTargetElementTrait.swift
```diff
@@ -95,21 +94,8 @@ private extension UIView {
...
-            if view.isMatch(for: target, on: \.tag) {
-                weight += 1_000
-            }
-
-            if view.isMatch(for: target, on: \.description) {
-                weight += 100
-            }
-
+        let weightedMatches = views.map { view -> (AppcuesViewElement, Int) in
+            let weight = view.selector?.evaluateMatch(for: selector) ?? 0
```
iujames, line 98: in theory, `strategy.findMatches(...)` could return some structure that included the weights as well, since it is checking it already to find matches. I kept the interface simpler to start and just returned the views and then have to re-check the weights here to find the best match, in this internal implementation detail.

Sources/AppcuesKit/Presentation/Public/AppcuesViewElement.swift
```diff
@@ -0,0 +1,92 @@
...
+@objc
+public class AppcuesViewElement: NSObject, Encodable {
+    /// Auto-generated unique ID for the view.
+    let id = UUID().appcuesFormatted
+
+    /// The x-coordinate for view position, with origin in the upper-left corner.
+    let x: CGFloat
+
+    /// The y-coordinate for view position, with origin is in the upper-left corner.
+    let y: CGFloat
```
mmaatttt, line 26: Perhaps specify these values are in screen coordinates, not relative to the parent?

mmaatttt (approved):
This is sweet! In particular, really like how the `AppcuesTargetElementTrait` has shaped up.

A bunch of small comments, but super excited for this.

Sources/AppcuesKit/Appcues.swift
iujames: sounds good, I'm on board with using the `static` here, seems appropriate in this case.

Sources/AppcuesKit/Appcues+Config.swift
```diff
@@ -173,5 +175,22 @@ public extension Appcues {
...
+        /// experiences. It is also used to control how elements are found and used during rendering of targeted element
+        /// experiences, such as anchored tooltips for example.
+        ///
+        /// If no strategy is supplied in the configuration, the SDK will use a default strategy for native iOS UIKit applications.
+        ///
+        /// - Parameter strategy: The strategy implementation to use for element targeted experiences.
+        /// - Returns: The `Configuration` object.
+        @discardableResult
+        @objc
+        public func elementTargeting(_ strategy: AppcuesElementTargeting) -> Self {
```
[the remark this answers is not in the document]
iujames: True, that is simpler - will update.

Sources/AppcuesKit/Presentation/Debugger/ScreenCapturing/UIKitElementTargeting.swift
```diff
@@ -0,0 +1,71 @@
...
+//
+
+import UIKit
+
+// UIKit selector implementation that uses accessibilityIdentifier,
+// accessibiltyLabel, and tag data from the underlying UIView to identify elements.
+internal class UIKitElementSelector: AppcuesElementSelector {
+
+    let accessibilityIdentifier: String?
+    let accessiblityLabel: String?
```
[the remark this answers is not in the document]
iujames: yes thank you. Interestingly, the base class already has `accessibilityLabel` since it inherits from NSObject, which adds that. I also discovered that we need to implement Encodable directly in the derived type to make inherited Encodable work as expected (`CodingKeys` and `encode(to encoder: Encoder)`)

in general, having to make this a derived class with NSObject in the hierarchy just for `@objc` compat is awkward, but I've yet to find a better way.

Sources/AppcuesKit/Presentation/Debugger/ScreenCapturing/UIKitElementTargeting.swift
```diff
@@ -0,0 +1,71 @@
...
+        return weight
+    }
+}
+
+// UIKit implementation of element targeting that captures the UIView hierarchy for the current UIWindow,
+// and identifies applicable views with a UIKitElementSelector.
+@available(iOS 13.0, *)
+internal class UIKitElementTargeting: AppcuesElementTargeting {
+    func captureLayout() -> AppcuesViewElement? {
+        return UIApplication.shared.windows.first { !$0.isAppcuesWindow }?.captureLayout()
```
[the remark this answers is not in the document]
iujames: I actually had it that way, restructured, and flip-flopped on it 😂 But yeah, I think it does make sense and I will move it back. Thanks for the nudge.

Sources/AppcuesKit/Presentation/Debugger/ScreenCapturing/UIKitElementTargeting.swift
```diff
@@ -32,6 +31,7 @@ internal class UIKitElementSelector: AppcuesElementSelector {
 
         super.init()
 
+        // note: accessibilityLabel is inhertied from base type NSObject and used here
```
mmaatttt, line 34: `inhertied`

Sources/AppcuesKit/Presentation/Public/AppcuesViewElement.swift
```diff
@@ -74,12 +76,13 @@ internal extension AppcuesViewElement {
     // Used internally by the SDK to find all views, recursively, that have at least a
     // partial match with the given selector. Used to find the position of an element
     // targeted experience.
-    func viewsMatchingSelector(_ target: AppcuesElementSelector) -> [AppcuesViewElement] {
-        var views: [AppcuesViewElement] = []
+    func viewsMatchingSelector(_ target: AppcuesElementSelector) -> [(AppcuesViewElement, Int)] {
```
mmaatttt, line 79: My personal preference is to name secondary, ambiguous (eg "what's the Int at first glance?") tuple elements, so I'd say
`func viewsMatchingSelector(_ target: AppcuesElementSelector) -> [(AppcuesViewElement, weight: Int)] {` is clearer.

The name doesn't have to be used anywhere else (though it can be used at the call site for additional clarity), but it makes it a bit clearer to read the intent.

Could do `(view: AppcuesViewElement, weight: Int)` too if we really want `AppcuesTargetElementTrait` to be more readable.

Sources/AppcuesKit/Presentation/Traits/Appcues/AppcuesTargetElementTrait.swift
```diff
@@ -82,34 +82,46 @@ internal class AppcuesTargetElementTrait: BackdropDecoratingTrait {
...
-                return maxItems[0].0
-            }
+        // if this has produced a single most distinct result, use it
+        if maxWeightViews.count == 1 {
+            return maxWeightViews[0]
         }
 
         // otherwise, this selector was not able to find a distinct match in this view
-        throw TraitError(description: "multiple non-distinct views (\(views.count)) matched selector \(config.selector)")
+        throw TraitError(description: "multiple non-distinct views (\(weightedViews.count)) matched selector \(config.selector)")
```
mmaatttt: I like this format better since throwing from a guard is the pattern used here and then we have the return at the end.

```swift
guard maxWeightViews.count == 1 else {
    // this selector was not able to find a distinct match in this view
    throw TraitError(description: "multiple non-distinct views (\(weightedViews.count)) matched selector \(config.selector)")
}

// if this has produced a single most distinct result, use it
return maxWeightViews[0]
``

mmaatttt (approved):
A couple more nits

Sources/AppcuesKit/Presentation/Debugger/ScreenCapturing/UIKitElementTargeting.swift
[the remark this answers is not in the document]
mmaatttt: Oh I wouldn't have expected that. Interesting

Sources/AppcuesKit/Presentation/Debugger/ScreenCapturing/UIKitElementTargeting.swift
iujames: 😅 and I even had my spell check on, doh. Thank you.

Sources/AppcuesKit/Presentation/Public/AppcuesViewElement.swift
iujames: good call, updating to use `(view: AppcuesViewElement, weight: Int)` and usage in `AppcuesTargetElementTrait` to maximize readability.

Sources/AppcuesKit/Presentation/Traits/Appcues/AppcuesTargetElementTrait.swift
iujames: yep, good call for consistency.

PR #371, opened 2023-03-23
Fix spelling errors

mmaatttt:
Now that we realized we can use spell checking in Xcode, I went a fixed all the typos at once.


iujames (approved):
_just a few_ 😅 thank you!

PR #377, opened 2023-04-04
Update docs for 2.0

mmaatttt:
Doc changes should be self explanatory.

I updated the skippable trait for readability because the long function was causing a well-justified lint warning.

I also modified the `AppcuesStepTransitionAnimationTrait` to use String value in the metadata so it can be read by third party traits since the animation block functions that read those values are internal only. This change makes it possible for third party traits to use the animation values.


Sources/AppcuesKit/AppcuesKit.docc/Traits.md
```diff
@@ -4,12 +4,18 @@ Experiences displayed by the Appcues iOS SDK are customizable and extensible via
 
 ## Overview
 
-An ``ExperienceTrait`` modifies the how an entire experience, or a particular step in an experience is displayed. A trait has capabilities that modify the way an experience is displayed to the user.
+An ``ExperienceTrait`` modifies the how an entire experience, or a particular step in an experience is displayed. A trait of a specific _type_ has _capabilities_ that modify the way an experience is displayed to the user. Furthermore a trait can be applied at multiple _levels_ which determines which parts of the experience are impacted by the trait instance.
```
mmaatttt, line 7: I've updated the doc a bit to be organized around Trait Types, Capabilities, Levels and I think it's easier to follow now. These three keywords are the critical concepts to understand, so hopefully this intro helps with that.

Sources/AppcuesKit/Presentation/Public/AppcuesElementTargeting.swift
```diff
@@ -10,8 +10,7 @@ import Foundation
 
 /// A type that describes an element targeting strategy for views in the application UI.
 ///
-/// The SDK provides a default implementation for element targeting, but it can be overridden at initialization of the SDK
-/// using the `Appcues.Config` option for `elementTargeting`.
+/// The default implementation for element targeting can be overridden by setting ``Appcues/elementTargeting``.
```
iujames, line 13: good catch, thank you

iujames (approved):
I think the Trait doc reads really well! Really helpful stuff there and hopefully of use outside of our group at some point as well. I'm sure already on your list, but note to also pass through here with updates when the `Appcues*` prefix is added to public types.

mmaatttt:
> I'm sure already on your list, but note to also pass through here with updates when the Appcues* prefix is added to public types.

Yep, PR for that is coming shortly.

PR #379, opened 2023-04-05
Add test harness that implements the breadth of the public interface

mmaatttt:
I mentioned wanting to do this a while back, but now is a good time as we prep for 2.0. This tests everything that's currently `public` in the SDK, with the idea that changing those things would cause the test case to fail to compile and act as a reminder that we require a major version bump.

When I update this for 2.0, it'll be a good way to create a migration guide when for example, this was removed:
```swift
_ = Font.Design.allCases
_ = Font.Design.default.description
_ = Font.Weight.allCases
_ = Font.Weight.regular.description
```

This test class obviously won't notify us about new public API additions, so we'll need to remember to add them.

In the process of creating this I noticed that we have a public `TraitError` with the intent that a custom trait could throw an instance, but there's no public init, so that won't work. We can make the init public or make `TraitError` internal and require 3rd party traits to have their own `Error` objects. In light of our documentation, I think making a public init is the right move.

### DIFF
--- a/Tests/AppcuesKitTests/PublicAPITests.swift
+++ b/Tests/AppcuesKitTests/PublicAPITests.swift
@@ -1,0 +1,247 @@
+//
+//  PublicAPITests.swift
+//  AppcuesKitTests
+//
+//  Created by Matt on 2023-04-05.
+//  Copyright © 2023 Appcues. All rights reserved.
+//
+
+import XCTest
+import SwiftUI
+import AppcuesKit
+
+/// This test class exists to verify the full breadth of the public API contract. It does not assert any specific results.
+/// If this test class fails to compile, that indicates a breaking change in the public API.
+class PublicAPITests: XCTestCase {
+
+    func testAPI() throws {
+        let config = Appcues.Config(accountID: "12345", applicationID: "abc")
+            .logging(true)
+            .apiHost(URL(string: "localhost")!)
+            .sessionTimeout(3600)
+            .activityStorageMaxSize(25)
+            .activityStorageMaxAge(3600)
+            .urlSession(URLSession.shared)
+            .anonymousIDFactory({ UUID().uuidString })
+            .additionalAutoProperties(["test": "value"])
+            .enableUniversalLinks(true)
+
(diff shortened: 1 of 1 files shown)


Tests/AppcuesKitTests/PublicAPITests.swift
```diff
@@ -0,0 +1,247 @@
...
+            .apiHost(URL(string: "https://api.appcues.net")!)
+            .sessionTimeout(3600)
+            .activityStorageMaxSize(25)
+            .activityStorageMaxAge(3600)
+            .urlSession(URLSession.shared)
+            .anonymousIDFactory({ UUID().uuidString })
+            .additionalAutoProperties(["test": "value"])
+            .enableUniversalLinks(true)
+
+        let appcuesInstance = Appcues(config: config)
```
iujames, line 29: using *real* `Appcues` (and underlying `NetworkClient`) and the correct endpoint with a valid account `12345` means that these calls would actually send production data into that account, correct? Most of our tests use the `MockAppcues`and don't do any real networking - would that suffice here?

iujames (approved):

iujames:
> In the process of creating this I noticed that we have a public TraitError with the intent that a custom trait could throw an instance, but there's no public init, so that won't work. We can make the init public or make TraitError internal and require 3rd party traits to have their own Error objects. In light of our documentation, I think making a public init is the right move.

Sounds like it would probably rename to `AppcuesTraitError` with the other proposed change? Having a public init there sounds good to me.

Tests/AppcuesKitTests/PublicAPITests.swift
mmaatttt: Good point. I like the `localhost` idea, so I'll do that. `MockAppcues` would work, but it adds a layer of obfuscation on what we're trying to test here, so I'd prefer to not use that.

mmaatttt:
> Sounds like it would probably rename to `AppcuesTraitError` with the other proposed change?

Yep, that's the plan

> Having a public init there sounds good to me.

👍

PR #365, opened 2023-03-16
3 little tooltip UI fixes

mmaatttt:
Take a look at each individual commit

PR #380, opened 2023-04-06
Fix tooltip corner radius when there’s a border

mmaatttt:
In #365 I did [🐛 Fix tooltip content clipping during initial layout](https://github.com/appcues/appcues-ios-sdk/pull/365/commits/dd0c1f9560080c8612e6c2d178570bbf87eb70b7), but that didn't work properly with border widths because the frame width/height for the inner mask was including the border width. Changed that and also made sure the initial width is 0 instead of 1.

|Fixed|Broken|
|-|-|
|Simulator Screen Shot - iPhone 14 Pro - 2023-04-06 at 11 28 46|Simulator Screen Shot - iPhone 14 Pro - 2023-04-06 at 11 28 14|

### DIFF
--- a/Sources/AppcuesKit/Presentation/Traits/Appcues/TooltipWrapperView.swift
+++ b/Sources/AppcuesKit/Presentation/Traits/Appcues/TooltipWrapperView.swift
@@ -55,7 +55,11 @@ internal class TooltipWrapperView: ExperienceWrapperView {
 
     private let maskLayer = CAShapeLayer()
     private let innerMaskLayer = CAShapeLayer()
-    private let borderLayer = CAShapeLayer()
+    private let borderLayer: CAShapeLayer = {
+        let layer = CAShapeLayer()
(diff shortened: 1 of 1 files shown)


Sources/AppcuesKit/Presentation/Traits/Appcues/TooltipWrapperView.swift
```diff
@@ -55,7 +55,11 @@ internal class TooltipWrapperView: ExperienceWrapperView {
 
     private let maskLayer = CAShapeLayer()
     private let innerMaskLayer = CAShapeLayer()
-    private let borderLayer = CAShapeLayer()
+    private let borderLayer: CAShapeLayer = {
+        let layer = CAShapeLayer()
+        layer.lineWidth = 0
```
iujames, line 60: was curious why this was needed, I see this defaults to one 👍

iujames (approved):

PR #381, opened 2023-04-06
Fix unwanted animation on first tooltip step

mmaatttt:
This was tricky because the animating effect that we don't want is actually the same as animating from the no-target-bottom-sheet scenario, which is something we support intentionally. I needed to find a way to differentiate between these cases. ([slack ref](https://appcues.slack.com/archives/C031RLGS4F8/p1680101315978779))

The best (only) approach I could find was handling it in the trait itself. It's not possible to differentiate the two scenarios in the view, because they look exactly the same.

What this PR does is add a non-animating metadata handler initially, and once we've seen a tooltip frame, then we switch to an animating handler. This way it's able to handle the initial case while also animating properly to a non-positioned tooltip.


Sources/AppcuesKit/Presentation/Traits/TraitMetadataDelegate.swift
```diff
@@ -43,8 +43,10 @@ public class TraitMetadataDelegate: NSObject {
 
     public func registerHandler(for key: String, animating: Bool, observer: @escaping (TraitMetadata) -> Void) {
         if animating {
+            nonAnimatingSubscribers.removeValue(forKey: key)
```
iujames, line 46: seems like this would always be the desired automatic behavior when re-registering the same key. Only thought is maybe the ever so slightly more DRY version of calling `removeHandler(for: key)` once at the top of the function? not important though.

iujames (approved):

Sources/AppcuesKit/Presentation/Traits/TraitMetadataDelegate.swift
mmaatttt: Updated to use `removeHandler` while rebasing 👍

PR #382, opened 2023-04-10
Move ExperienceTraitLevel to ExperiencePluginConfiguration

mmaatttt:
Way back in https://github.com/appcues/appcues-ios-sdk/pull/231 we added `ExperienceTraitLevel` which gets passed in the `ExperienceTrait.init`. This is fine as is, but it’s really not something that gets used much (it was designed only for one specific trait), so I was thinking we could move it into `ExperiencePluginConfig` as a property. This in theory has a few benefits, none of which are huge, but add up to a nice improvement:

1. `ExperienceAction` inits could now use this (not that we have a real good use case for it, but it would allow a trait to know if it’s being used at the group level)
2. The trait init is simpler and we have a pattern to include other new things in the plugin config object instead of the trait init which creates more flexibility for us to add things (including `internal`-only things) without breaking changes to the init interface.
3. `ExperienceTraitLevel` can be renamed to `ExperiencePluginConfig.Level` which removes it from the list of public things we’re exposing at the root level of the API, which isn’t that meaningful, but I like keeping that list simple where possible.

This is a breaking change 💥 so looking to do it now before we lock in 2.0

### DIFF
--- a/Sources/AppcuesKit/Presentation/Actions/ActionRegistry.swift
+++ b/Sources/AppcuesKit/Presentation/Actions/ActionRegistry.swift
@@ -65,9 +65,9 @@ internal class ActionRegistry {
 
     /// Enqueue an array of experience action data models to be executed. This version is for non-interactive action execution,
     /// such as actions that execute as part of the navigation to a step.
-    func enqueue(actionModels: [Experience.Action], completion: @escaping () -> Void) {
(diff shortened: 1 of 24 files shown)


iujames (approved):
very nice! a good simplification of the integration point here, and future flexibility as you noted.

PR #387, opened 2023-04-12
Don’t deploy SDK docs for pre-release versions

mmaatttt:
The published docs should always reflect the latest stable release.

And `bundle update` for the latest dependabot.


iujames (approved):
nice update

PR #394, opened 2023-04-26
Add target interaction trait to control interactions with target rectangle

iujames:
This adds an implementation of `@appcues/target-interaction` (naming tbd, [spec PR](https://github.com/appcues/appcues-mobile-experience-spec/pull/175)). This trait can be used to create a transparent UIView that lays over the target rectangle, from metadata, and captures taps and long presses. The trait can supply actions to take on those interactions, using standard experience action format. If no actions are provided, the trait simply captures those interactions and prevents them from flowing through to the backdrop - meaning it would not dismiss the flow, for example, if the skippable trait was otherwise dismissing on other backdrop taps.

One notable update is that `AppcuesExperiencePluginConfiguration` now includes the `Appcues` instance. This was made available here so that traits can have access to use this instance if needed. This new trait will use the instance to get the `ActionRegistry` and execute actions, when necessary.

### DIFF
--- a/Sources/AppcuesKit/Presentation/Public/Plugins/AppcuesExperiencePluginConfiguration.swift
+++ b/Sources/AppcuesKit/Presentation/Public/Plugins/AppcuesExperiencePluginConfiguration.swift
@@ -28,9 +28,13 @@ public class AppcuesExperiencePluginConfiguration: NSObject {
     /// The context where the plugin was defined.
     public let level: Level
 
-    init(_ decoder: PluginDecoder, level: Level) {
+    /// The instance of the Appcues SDK where the plugin is being applied.
+    public weak var appcues: Appcues?
+
+    init(_ decoder: PluginDecoder, level: Level, appcues: Appcues?) {
         self.decoder = decoder
         self.level = level
+        self.appcues = appcues
     }
 
     /// Returns a value of the type you specify, decoded from a JSON object.
--- a/Sources/AppcuesKit/Presentation/Traits/Appcues/AppcuesTargetBehaviorTrait.swift
+++ b/Sources/AppcuesKit/Presentation/Traits/Appcues/AppcuesTargetBehaviorTrait.swift
@@ -1,0 +1,108 @@
+//
+//  AppcuesTargetBehaviorTrait.swift
+//  AppcuesKit
+//
+//  Created by James Ellis on 4/25/23.
+//  Copyright © 2023 Appcues. All rights reserved.
+//
+
+import UIKit
+
+@available(iOS 13.0, *)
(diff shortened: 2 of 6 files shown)


Sources/AppcuesKit/Presentation/Public/Plugins/AppcuesExperiencePluginConfiguration.swift
```diff
@@ -28,9 +28,13 @@ public class AppcuesExperiencePluginConfiguration: NSObject {
     /// The context where the plugin was defined.
     public let level: Level
 
-    init(_ decoder: PluginDecoder, level: Level) {
+    /// The instance of the Appcues SDK where the plugin is being applied.
+    public weak var appcues: Appcues?
```
iujames, line 32: discussion ongoing, but now that the Appcues instance is here, we do not technically need to also include in calls to AppcuesExperienceAction `execute(inContext, completion)`

Sources/AppcuesKit/Presentation/Traits/Appcues/AppcuesTargetBehaviorTrait.swift
```diff
@@ -0,0 +1,108 @@
...
+    private func didTapTarget() {
+        guard !tapActions.isEmpty, let actionRegistry = appcues?.container.resolve(ActionRegistry.self) else {
+            return
+        }
+
+        actionRegistry.enqueue(
+            actionModels: tapActions,
+            level: .step,
+            interactionType: "Target Tapped",   // ??
+            viewDescription: "Target Rectangle" // ??
```
iujames, line 91: I'm speculating that including the `appcues:v2:step_interaction` analytics, just like button interactions, is desired here. I'm not totally sure though. Also not sure what the values for interactionType or viewDescription should be in this case, so I put in these initial values to start.

Sources/AppcuesKit/Presentation/Public/Plugins/AppcuesExperiencePluginConfiguration.swift
iujames: decision is to remove the appcues instance from the execute function signature, another change coming on top of this for that update.

Sources/AppcuesKit/Presentation/Traits/Appcues/AppcuesTargetBehaviorTrait.swift
iujames: moving forward with this step_interaction approach.

andretortolano (approved):

mmaatttt (approved):

PR #395, opened 2023-04-27
Remove appcues parameter from AppcuesExperienceAction.execute

mmaatttt:
This is a 💥 breaking change so the 2.0 migration guide has been updated accordingly.

The most interesting bit is the post experience actions since their `init` from the `Experience` extension now require the `Appcues` instance. Rather than save that reference in the `ExperienceStateMachine`, I’ve updated it to `var postExperienceActionFactory: ((Appcues?) -> [AppcuesExperienceAction])`, which instead of returning the actual action _instances_, returns a function that generates the actions given an `Appcues` instance. Then in the `ActionRegistry`, we can give it that instance, consistent with the way we do the regular action inits.

This PR also updates the `@appcues/skippable` trait to use the `Appcues` reference to call the `ExperienceRenderer` to dismiss rather than hackily relying on `UIViewController.dismiss` which works for the current PresentingTraits that we have but isn't guaranteed to work for all.


Sources/AppcuesKit/Presentation/Actions/ActionRegistry.swift
```diff
@@ -72,11 +72,17 @@ internal class ActionRegistry {
         execute(transformQueue(actionInstances), completion: completion)
     }
 
-    /// Enqueue an array of experience actions instance to be executed. This version is used for post-completion actions on an experience.
+    /// Enqueue an array of experience actions instances to be executed.
```
iujames, line 75: nit, but this one can now become private in the class, or inlined into its one usage spot on line 112

Sources/AppcuesKit/Presentation/Actions/Appcues/AppcuesCloseAction.swift
```diff
@@ -17,14 +17,20 @@ internal class AppcuesCloseAction: AppcuesExperienceAction {
 
     static let type = "@appcues/close"
 
+    private weak var appcues: Appcues?
+
     private let markComplete: Bool
 
     required init?(configuration: AppcuesExperiencePluginConfiguration) {
+        appcues = configuration.appcues
```
iujames, line 25: just thinking, alternative pattern would be to resolve the `experienceRender` right here and store a weak reference to that? then the execute would just be 
```
experienceRenderer?.dismissCurrentExperience(markComplete: markComplete) { _ in completion() }
```
any pros/cons either way that you can see? I suppose would still need some sort of guard on nil to ensure the completion was called. This seems fine, just thinking out loud. The `return completion()` syntax caught me off guard.

Sources/AppcuesKit/Presentation/Actions/Appcues/AppcuesSubmitFormAction.swift
```diff
@@ -16,14 +16,20 @@ internal class AppcuesSubmitFormAction: AppcuesExperienceAction, ExperienceActio
...
     required init?(configuration: AppcuesExperiencePluginConfiguration) {
+        self.appcues = configuration.appcues
+
         let config = configuration.decode(Config.self)
         self.skipValidation = config?.skipValidation ?? false
     }
 
-    func execute(inContext appcues: Appcues, completion: ActionRegistry.Completion) {
+    func execute(completion: ActionRegistry.Completion) {
+        guard let appcues = appcues else { return completion() }
```
iujames, line 31: in this case would you want to have this guard after the `defer` on line below and then just have the completion call in one spot?

Sources/AppcuesKit/Presentation/ExperienceRendering/ExperienceStateMachine.swift
```diff
@@ -216,7 +216,7 @@ extension ExperienceStateMachine {
         case navigateInContainer(ExperiencePackage, pageIndex: Int)
         case dismissContainer(ExperiencePackage, continuation: Action)
         case error(ExperienceError, reset: Bool)
-        case processActions([AppcuesExperienceAction])
+        case processActions((Appcues?) -> [AppcuesExperienceAction])
```
iujames, line 219: nice approach!

Sources/AppcuesKit/Presentation/Traits/Appcues/AppcuesSkippableTrait.swift
```diff
@@ -68,8 +74,11 @@ internal class AppcuesSkippableTrait: AppcuesContainerDecoratingTrait, AppcuesBa
     }
 
     @objc
-    private func didTapBackground() {
-        containerController?.dismiss(animated: true)
+    private func dismissExperience() {
+        guard let appcues = appcues else { return }
+
+        let experienceRenderer = appcues.container.resolve(ExperienceRendering.self)
+        experienceRenderer.dismissCurrentExperience(markComplete: false, completion: nil)
```
iujames, line 81: nice improvement

Tests/AppcuesKitTests/Actions/AppcuesCloseActionTests.swift
```diff
@@ -20,7 +20,7 @@ class AppcuesCloseActionTests: XCTestCase {
 
     func testInit() throws {
         // Act
-        let action = AppcuesCloseAction(configuration: AppcuesExperiencePluginConfiguration(nil))
+        let action = AppcuesCloseAction(configuration: AppcuesExperiencePluginConfiguration(nil, appcues: appcues))
```
iujames, line 23: ah yes, my little shortcut of just using a default nil value for `appcues` in the test convenience init was not too helpful for you on these close actions 😬 thanks for cleaning up

iujames (approved):
nice clean up to get into 2.0!

Sources/AppcuesKit/Presentation/Actions/Appcues/AppcuesSubmitFormAction.swift
mmaatttt: Definitely an option. I did it like this for the sake of consistency with all the other actions. Happy to update it if you think it'd be better.

Sources/AppcuesKit/Presentation/Actions/Appcues/AppcuesCloseAction.swift
mmaatttt: Right. Both approaches are workable. I started off resolving the dependencies in the init but then I ended up doing them all with `Appcues` for the sake of consistency since some actions actually use the `Appcues` interface directly.

Not that it's a good reason, but referencing `Appcues` vs the dependencies also means fewer changes from the way it was before.

iujames: yep, this all seems good to me as-is, thanks for indulging my ramblings

Sources/AppcuesKit/Presentation/Actions/Appcues/AppcuesSubmitFormAction.swift
iujames: yeah, argument for consistency makes sense, sounds good to me.

PR #399, opened 2023-05-01
Fix incorrect color on backdrop with circular keyhole

mmaatttt:
> Gradient with locations close to 1.0 was failing to transition fully to the expected color. Doubling the size of the gradient and then moving the location stops to 0.5 fixes this.

For a circular keyhole with no `spreadRadius` we were still doing a tiny gradient (this enables animation between `spreadRadius` values), but the gradient stop position values are so close together at the edge of the gradient (0.995 and 1.0) it’s not getting all the way to the full opacity mask and the result is a lighter backdrop color than expected.

The solution here is to move the gradient stop positions to 50% (ie color locations at 0.4999995, 0.5), so I adjusted the gradient to be double the size but then have the locations at the 50% point instead of 100%.

Snapshot tests in https://github.com/appcues/appcues-mobile-experience-spec/pull/180


iujames (approved):
nice fix!

PR #403, opened 2023-05-10
Add tutorial for creating permission primers

mmaatttt:
DocC tutorial that walks through the in-app, builder, and studio targeting work necessary to create a permission primer flow (for camera access).


Sources/AppcuesKit/AppcuesKit.docc/Tutorials/PermissionPrimers.tutorial
```diff
@@ -0,0 +1,120 @@
...
+            @Step {
+                Create an extension on `AVAuthorizationStatus` to get a standardized, human-readable string for each possible status.
+                
+                @Code(name: "AVAuthorizationStatus+Description.swift", file: PermissionPrimers-01-02.swift)
+            }
+            
+            @Step {
+                Update your existing `Appcues.identify()` call to set a user property with the current camera permission status.
+                                
+                @Code(name: "MyAppAuth.swift", file: PermissionPrimers-01-03.swift, previousFile: PermissionPrimers-01-03-0.swift)
```
iujames, line 29: previousFile is how it knows which lines to highlight eh? slick!

Sources/AppcuesKit/AppcuesKit.docc/Tutorials/PermissionPrimers.tutorial
```diff
@@ -0,0 +1,120 @@
...
+                @Image(source: PermissionPrimers-02-02.png, alt: "Mobile permission primer flow")
+            }
+            
+            @Step {
+                Design your flow.
+                
+                Consider reviewing the [Human Interface Guidelines for Pre-alert screens](https://developer.apple.com/design/human-interface-guidelines/accessing-private-data#Pre-alert-screens).
+                
+                @Image(source: PermissionPrimers-02-02.png, alt: "Mobile permission primer flow")
+            }
```
iujames, line 84: duplicate step 3 here with same content of step 2

iujames (approved):
awesome! the tutorial form is amazing for iOS devs I think. We'll have to port to other platforms in some form.

one little detail below about a dupe step.

This keeps it pretty simple and straightforward, which I think is the right call. I think there is no concern about the case of going to settings and changing the permission status outside of the app, for this use case, correct? since the option will not appear in settings until it is asked?

Sources/AppcuesKit/AppcuesKit.docc/Tutorials/PermissionPrimers.tutorial
mmaatttt: Oops, thanks. Do you think the 30 minute estimate is reasonable? It's not super helpful since it's probably different people doing different sections, but...

iujames: I think 30 seems reasonable to read and find the right spots to adjust in the app and test it out. Fair point that the end designer of the flow would likely be someone else in the org later on down the line.

PR #404, opened 2023-05-12
Add pipeline orchestration to run snapshot tests in spec repo

mmaatttt:
This is the other half of https://github.com/appcues/appcues-mobile-experience-spec/pull/185 based on https://circleci.com/blog/pipeline-orchestration-circleback/

There's 3 new jobs here:

1. `trigger-spec-test-pipeline` where we want to run the unit tests alongside the snapshot tests, so we trigger the `appcues-mobile-experience-spec` pipeline (trying first for a branch name matching the one triggering and then falling back to `main`) with the name of the branch we want to test against.
2. `wait-for-spec-test-pipeline` an approval job that suspends this pipeline until an approval action is triggered (in this case by the `appcues-mobile-experience-spec` pipeline.
3. `check-status-of-spec-test-pipeline` where we use the saved ID of the triggered job to get its status, and if successful get the code coverage report to upload to CodeCov.

I've removed the existing `validate_code` job since the unit tests are run on the remote job, and then since `pr_validation` and `main_verification` were doing the same thing, I replaced them with a new name `branch_validation`.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -187,24 +187,26 @@ jobs:
         - persist_to_workspace:
             root: .
             paths: 
-              - pipeline.txt  
+              - pipeline.txt
   check-status-of-spec-test-pipeline:
     docker: 
       - image: cimg/base:current
     resource_class: small 
     steps:
+      # checkout so the source files are here for the codecov upload
+      - checkout
       - attach_workspace:
           at: workspace
       - run:
           name: Check triggered workflow status
           command: |
             triggered_pipeline_id=$(cat workspace/pipeline.txt)
-            created_workflow_status=$(curl --request GET \
+            curl --request GET \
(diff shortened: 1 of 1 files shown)


iujames (approved):
Very cool! 

I think most of this should translate pretty directly to the Android tests in the spec repo. That test project pulls in the Appcues SDK from [relative path](https://github.com/appcues/appcues-mobile-experience-spec/blob/main/verification-suites/android/settings.gradle#L18). So as long as we get the branches lined up and the files in the right relative path, I think we can access all the tests and run just like was done here. I have not looked into how the Jacoco reporting for Codecov would work on the spec repo yet though ([main SDK reference](https://github.com/appcues/appcues-android-sdk/blob/main/appcues/codecov.gradle))

PR #405, opened 2023-05-16
Fix format of survey form response field ID

mmaatttt:
Found in https://appcues.slack.com/archives/C03JYPH1NJZ/p1684250936524709

I tried to survey the rest of the codebase to make sure we don't have this issue anywhere else, but it's a bit tricky to be sure. We've got it correct already in places like https://github.com/appcues/appcues-ios-sdk/blob/fab91923f6ad9624164c7d9abdd5bf6d65ac620c/Sources/AppcuesKit/Data/Models/Activity.swift#L58

### DIFF
--- a/Sources/AppcuesKit/Data/Extensions/UUID+Custom.swift
+++ b/Sources/AppcuesKit/Data/Extensions/UUID+Custom.swift
@@ -8,7 +8,7 @@
 
 import Foundation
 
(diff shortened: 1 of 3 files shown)


iujames (approved):

iujames:
relates to [sc-52533]

PR #408, opened 2023-05-23
Fix deep link launching debugger when app is backgrounded

mmaatttt:
I'm in the process of creating UI tests for launching the debugger (and screen capture) using deep links and realized it's not working properly for one of the scenarios. If the client app isn't in the foreground, there's a mismatch checking the state and so the debugger launching fails.

There's also the case where the app is foregrounded which isn't really possible on a device launching from a QR code in the camera app, but is often the case when launching the deep link on a simulator, which I think is why we haven't noticed it not working. Additionally, the experience preview deep link always works because the way it presents the experience is different than the logic used in the debugger.

UI tests to follow in the spec repo.


iujames (approved):
nice catch! definitely would have bit us in the upcoming screen capture QR code work.

iujames (approved):

PR #410, opened 2023-05-30
Add View modifier to support SwiftUI element targeting

mmaatttt:
The modifier adds a background `AppcuesTargetView` `UIView` which has the same frame as a the tagged view, allowing it to be targeted.

Also updating the default Name of captured screens to be nicer for SwiftUI views and removed the (now) redundant timestamp.

### DIFF
--- a/Sources/AppcuesKit/Presentation/Debugger/ScreenCapturing/UIKitElementTargeting.swift
+++ b/Sources/AppcuesKit/Presentation/Debugger/ScreenCapturing/UIKitElementTargeting.swift
@@ -12,6 +12,7 @@ import UIKit
 // accessibilityLabel, and tag data from the underlying UIView to identify elements.
 internal class UIKitElementSelector: AppcuesElementSelector {
     private enum CodingKeys: String, CodingKey {
+        case appcuesID
         case accessibilityIdentifier
         case accessibilityLabel
         case tag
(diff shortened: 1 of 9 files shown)


Sources/AppcuesKit/Presentation/Debugger/ScreenCapturing/UIKitElementTargeting.swift
```diff
@@ -44,6 +47,10 @@ internal class UIKitElementSelector: AppcuesElementSelector {
         // weight the selector property matches by how distinct they are considered
         var weight = 0
 
+        if appcuesID != nil && appcuesID == other.appcuesID {
+            weight += 10_000
```
mmaatttt, line 51: I think using 10000 here is ok? Or do we not want it to match `accessibilityIdentifier`?

iujames (approved):
such a simple and clever way to enable this, love it!

Sources/AppcuesKit/Presentation/Debugger/ScreenCapturing/UIKitElementTargeting.swift
andretortolano: I was wondering if later we could change that values to constant values that would work with bit operations, Ive never done it but you know like masks, where you can combine masks to have a final value, maybe that would be a way to go here, also the weight values are not as important since when we add new selector types we can always change the weight they have. its not something we need to keep set in stone forever right?

PR #411, opened 2023-06-01
Ensure experiences are dismissed before screen capture

mmaatttt:
The debugger now has a dependency on the `ExperienceRenderer` to be able to dismiss an experience if needed before performing a screen capture:

iOS SDK Dependency Graph-2

Also took care of sc-53273 and rename the target-element trait, as well as added a toast message in place of a silent failure if the screen capture fails.

### DIFF
--- a/Sources/AppcuesKit/Presentation/Debugger/UIDebugger.swift
+++ b/Sources/AppcuesKit/Presentation/Debugger/UIDebugger.swift
@@ -153,9 +153,7 @@ extension UIDebugger: DebugViewDelegate {
               let window = UIApplication.shared.windows.first(where: { !$0.isAppcuesWindow }),
               let screenshot = window.screenshot(),
               let layout = Appcues.elementTargeting.captureLayout() else {
-
-            // show error?
-
+            debugViewController?.showCaptureFailure()
             return
         }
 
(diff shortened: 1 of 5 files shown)


Sources/AppcuesKit/Presentation/Debugger/UIDebugger.swift
```diff
@@ -148,14 +150,19 @@ extension UIDebugger: DebugViewDelegate {
         }
     }
 
-    private func captureScreen(authorization: Authorization) {
+    private func captureScreen(authorization: Authorization, dismissed: Bool = false) {
```
iujames, line 153: just curious why the `dismissed` param is needed, if the guard is already checking the `experienceRenderer.getCurrentStepIndex() == nil`, which would get set to nil after the `dismissCurrentExperience` completes?

iujames (approved):

Sources/AppcuesKit/Presentation/Debugger/UIDebugger.swift
mmaatttt: Good question. I started with the flag and then added the current step check afterwards to skip the dismiss call if it's unneeded, but it didn't cross my mind that the flag became redundant. I'll update it!

Sources/AppcuesKit/Presentation/Debugger/UIDebugger.swift
```diff
@@ -148,14 +150,19 @@ extension UIDebugger: DebugViewDelegate {
...
+        guard dismissed || experienceRenderer.getCurrentStepIndex() == nil else {
```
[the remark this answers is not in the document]
mmaatttt: Yep, good point.

PR #414, opened 2023-06-07
Add identify custom properties to subsequent event identity properties

mmaatttt:
When a user is identified with the mobile SDK, if there are custom properties that are included in that identification, those properties should be persisted and included in and events following that identification until there is a new identification or session.

Note that the first commit here refactors the tests a bit to be more readable, but nothing changed how the tests worked. The last test case `testProfilePropertiesOnSubsequentEvent` is the new one.

### DIFF
--- a/Sources/AppcuesKit/AppcuesKit.docc/Identifying.md
+++ b/Sources/AppcuesKit/AppcuesKit.docc/Identifying.md
@@ -8,6 +8,8 @@ In order to target content to the right users at the right time, you need to ide
 
 ``Appcues/identify(userID:properties:)``
 
+It is recommended that the application identify a user at moments such as sign in, and also when the app starts up on a cold launch with existing log in credentials. This will ensure that any user properties tied to this user can be kept accurately up to date.
+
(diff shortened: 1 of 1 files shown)


iujames:
maybe we should stick a note in here https://github.com/appcues/appcues-ios-sdk#identifying-users

something like: "It is recommended that the application identify a user at moments such as sign in, and also when the app starts up on a cold launch with existing log in credentials. This will ensure that any user properties tied to this user can be kept accurately up to date."

mmaatttt:
> maybe we should stick a note in here https://github.com/appcues/appcues-ios-sdk#identifying-users
> 
> something like: "It is recommended that the application identify a user at moments such as sign in, and also when the app starts up on a cold launch with existing log in credentials. This will ensure that any user properties tied to this user can be kept accurately up to date."

@iujames maybe in Sources/AppcuesKit/AppcuesKit.docc/Identifying.md? It feels more like an edge case note, so I'd rather put it in the detailed doc rather than the main README.

PR #415, opened 2023-06-14
Refactor ExperienceRenderer to support multiple ExperienceStateMachine instances for non-modal Experiences

mmaatttt:
This is all very close to the original POC (#237), where `ExperienceRenderer` has a dictionary of non-modal state machines. All the `ExperienceRenderer` methods are updated to apply to the provided state machine and then there's new methods to get the correct state machine for a given `experienceID` (which is actually the `Experience.instanceID`—see note 1 below).

The general blast radius of changes is as follows:
1. `ExperienceRenderer` now creates/manages multiple state machines
2. Things that call `ExperienceRenderer` (in particular traits/actions), need to know the `experienceID` to update the rendering of the correct experience
3. `AppcuesExperiencePluginConfiguration` needs to provide the `experienceID` to traits/actions
4. `ActionRegistry` and `TraitRegistry` need to provide the `experienceID` to `AppcuesExperiencePluginConfiguration`
5. `ExperienceStepViewModel` needs to provide the `experienceID` to `ActionRegistry`
6. `TraitComposer` gets the `experienceID` to provide to `ExperienceStepViewModel`

## Notes

1. I've introduced the lightweight `InstanceID` struct that just wraps a UUID for usage with `Experience.instanceID`. This value gets passed everywhere that we're using the `instanceID` (which now includes traits/actions/etc) and provides some compile time safety to ensure we don't actually pass the `Experience.id` or some other value. Without this, we would've had `let experienceID: String?` many different places which I found creating too much cognitive overload to remember that it's actually the `Experience.instanceID` not the `Experience.ID`.
2. The `UIDebugger` relies on a a quirk of how `ExperienceRenderer` treats a `nil` `experienceID`. Since the debugger doesn't know the `experienceID` of a modal experience, it passes `nil`, for which the helper `ExperienceRender.stateMachine(experienceID:)` returns the shared (ie modal) state machine to properly dismiss a modal experience. We might want the debugger to dismiss all (including non-modal experiences) before capturing a screen, or update the screen capture logic to stop when it gets to an embedded experience.
3. `AppcuesExperiencePluginConfiguration.instanceID` is internal since the value is meaningless outside the SDK, so there's no use case where a custom plugin would need it.


iujames (approved):
brings back memories! looking really clean, and good call on the `InstanceID` struct to handle that complexity in a safe way.

It will be interesting to see how the renderer evolves with all of the cache concepts we've discussed, but this foundation gets things well on the way!

PR #417, opened 2023-06-15
Add auto-generated tab selector property for standard UITabBarButton items

iujames:
This will create an index based selector for typical UITabBarViewController usage like
```js
"selector" : {
    "tab" : "tab[0]"
}
```
for the first tab in the view, for instance - zero based indices. This will also show up as the `displayName` for the view, if there are no higher priority selector props defined (`appcuesID` or `accessibilityIdentifier`). 

This will allow for customers to avoid having to do [workarounds like we do in the sample app](https://github.com/appcues/appcues-ios-sdk/blob/main/Examples/DeveloperCocoapodsExample/CocoapodsExample/HomeTabBarController.swift#L15-L25) just to get accessibilityIdentifiers into the common tab views (not supported by default). And will enable usage in pure SwiftUI apps using tabs.


Sources/AppcuesKit/Presentation/Debugger/ScreenCapturing/UIKitElementTargeting.swift
```diff
@@ -16,21 +16,30 @@ internal class UIKitElementSelector: AppcuesElementSelector {
...
         case tag
+        case tab
     }
 
     let accessibilityIdentifier: String?
     let tag: String?
     let appcuesID: String?
-
-    init?(appcuesID: String?, accessibilityIdentifier: String?, accessibilityLabel: String?, tag: String?) {
+    let tab: String?
```
mmaatttt, line 25: This is more ramblings than a concrete suggestion, but...

I don't know if there'll be other UIKit things we'll one day want to automatically add selectors for, but it could get unwieldy quickly if we add a new named value like `tab` every time. We could maybe do a more generic name here (idk what) and we could reuse across mutually exclusive UIKit types. Like if wanted to automatically tag UIBarButtonItem's or UISwitch's or something...

mmaatttt (approved):
Nice!

Sources/AppcuesKit/Presentation/Debugger/ScreenCapturing/UIKitElementTargeting.swift
iujames: Yeah, good call. Pondering names.

viewData, viewInfo, descriptor, generated, autoTag...

mmaatttt: I like `autoTag`, maybe because it's in the same vein as our auto-properties? Maybe `autoSelector`? (sounds like car shopping... 🙃)

iujames: Ok lets roll with `autoTag` then for now, seems decent enough.

PR #421, opened 2023-06-20
Add support for text scaling

iujames:
aka Dynamic Type accessibility

This approach would provide a config option to `.enableTextScaling(enabled)`. This option would then auto scale the font size given in the model to meet the users current desired text size scale. Since we can't map directly between our builder font sizes and the iOS TextStyle options for scaling, we use a simple mapping for 3 scale levels - effectively small / medium / large, to give a reasonable result.

This change sets this option false by default, and removes the previous accidental text scaling on custom fonts. Previously, custom fonts would scale and system fonts would not. Now they both scale only when this option is enabled.

This example has a custom Lato-Black headline and then system text below. The OS setting is for a larger font size than the default.

| disabled | enabled |
| ---- | ---- |
| Screenshot 2023-06-20 at 3 30 09 PM | Screenshot 2023-06-20 at 3 30 30 PM |

### DIFF
--- a/Sources/AppcuesKit/Appcues+Config.swift
+++ b/Sources/AppcuesKit/Appcues+Config.swift
@@ -43,6 +43,8 @@ public extension Appcues {
 
         var enableUniversalLinks = true
 
+        var enableTextScaling = false
+
         /// Create an Appcues SDK configuration
         /// - Parameter accountID: Appcues Account ID - a string containing an integer, copied from the Account settings page in Studio.
         /// - Parameter applicationID: Appcues Application ID - a string containing a UUID,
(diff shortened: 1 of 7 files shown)


Sources/AppcuesKit/Presentation/Traits/TraitComposer.swift
```diff
@@ -18,12 +18,14 @@ internal class TraitComposer: TraitComposing {
...
     private let traitRegistry: TraitRegistry
     private let actionRegistry: ActionRegistry
+    private let config: Appcues.Config
     private let notificationCenter: NotificationCenter
 
     init(container: DIContainer) {
         traitRegistry = container.resolve(TraitRegistry.self)
         actionRegistry = container.resolve(ActionRegistry.self)
         notificationCenter = container.resolve(NotificationCenter.self)
+        config = container.resolve(Appcues.Config.self)
```
mmaatttt, line 28: I'll add the dependency to the whimsical graph

Sources/AppcuesKit/Appcues+Config.swift
```diff
@@ -173,5 +175,21 @@ public extension Appcues {
...
+        /// When this option is enabled, Appcues content rendered by the SDK will support Dynamic Type, and adjust
+        /// to the user's preferred reading size, set in system settings.
+        ///
+        /// The default value for this configuration is `false`.
+        ///
+        /// - Parameter enabled: Whether text scaling is enabled.
+        /// - Returns: The `Configuration` object.
+        @discardableResult
+        @objc
+        public func enableTextScaling(_ enabled: Bool) -> Self {
```
mmaatttt, line 190: You can update the `PublicAPITests` to include this new function.

mmaatttt (approved):
Overall approach looks good! Having the config option is nice.

Sources/AppcuesKit/Presentation/Traits/TraitComposer.swift
iujames: thank you!

PR #424, opened 2023-06-22
Add docs for configuring views for element targeting

iujames:


Sources/AppcuesKit/AppcuesKit.docc/AnchoredTooltips.md
```diff
@@ -0,0 +1,51 @@
+# Configuring Views for Anchored Tooltips
+
+The Appcues iOS SDK supports anchored tooltips targeting views built with `UIKit` or `SwiftUI`.
+
+Instrumenting your application views as described below allows the Appcues iOS SDK to create a mobile view selector. This selector is used by the Appcues Mobile Builder to create and target anchored tooltips. When a user qualifies for a flow, this selector is used to render the anchored tooltip content in the correct location.
+
+## Instrumenting UIKit Views
+
+The following `UIView` properties are used to identify elements, in order of precedence:
```
iujames, line 9: I could not see a way to link Apple class references like `UIView` to the DocC page directly with the double backtick syntax. In items below, I used web URL, which does seem to actually link directly into the doc page inside of the doc browser when available, but did not do this for every reference.

Sources/AppcuesKit/AppcuesKit.docc/AnchoredTooltips.md
```diff
@@ -0,0 +1,51 @@
...
+button.tag = 1234
+```
+
+These properties are also available in Xcode Interface Builder, used for Storyboard UI development. Click on the view and select the Identity inspector tab to see the Accessibility section and set related properties. The `tag` is also visible to edit on the Attributes inspector. These properties can also be set using the User Defined Runtime Attributes section of the Identity inspector for a selected element in Interface Builder.
+
+> Tip: For the common use case of `UITabBarController` and its associated `UITabBar` items, the Appcues iOS SDK will automatically generate index-based selectors for each item to use for targeting tooltips. No additional code is required in your application if you would like to target the items in a `UITabBar` in your application's screen by index.
+
+## Instrumenting SwiftUI Views
+
+iOS Applications using SwiftUI can also identify `View` elements. A `View` extension `appcuesView(identifier:)` is provided by the Appcues iOS SDK to support this use case. The `identifier` String value must be unique on the screen where an anchored tooltip may be targeted.
```
iujames, line 32: Was also unable to find a way to link to public API docs for this `appcuesView(identifier:)` extension. This is an extension method on `View`, and perhaps does not generate DocC output?

Sources/AppcuesKit/AppcuesKit.docc/AnchoredTooltips.md
mmaatttt: Only symbols in the AppcuesKit module can be linked with double backticks.

Sources/AppcuesKit/AppcuesKit.docc/AnchoredTooltips.md
mmaatttt: It looks like Swift 5.9 and so maybe Xcode 14 might include the ability to do this, but it's not possible now.

https://forums.swift.org/t/enablement-of-docc-extension-support-as-an-opt-in-feature/62614/12

Sources/AppcuesKit/AppcuesKit.docc/AnchoredTooltips.md
```diff
@@ -0,0 +1,51 @@
...
+```swift
+let button = UIButton()
+button.accessibilityIdentifier = "btnSaveProfile"
+button.accessibilityLabel = "Save"
+button.tag = 1234
+```
+
+These properties are also available in Xcode Interface Builder, used for Storyboard UI development. Click on the view and select the Identity inspector tab to see the Accessibility section and set related properties. The `tag` is also visible to edit on the Attributes inspector. These properties can also be set using the User Defined Runtime Attributes section of the Identity inspector for a selected element in Interface Builder.
+
+> Tip: For the common use case of `UITabBarController` and its associated `UITabBar` items, the Appcues iOS SDK will automatically generate index-based selectors for each item to use for targeting tooltips. No additional code is required in your application if you would like to target the items in a `UITabBar` in your application's screen by index.
```
mmaatttt, line 28: It's cool how DocC formats the Tip. Looks great!

Sources/AppcuesKit/AppcuesKit.docc/AppcuesKit.md
```diff
@@ -44,6 +44,7 @@ A Swift library for sending user properties and events to the Appcues API and re
 - <doc:Traits>
 
 ### Element Targeting
```
mmaatttt, line 46: Thinking this heading maybe should be "Anchored Tooltips"? (my bad)

Actually, maybe move this doc reference up under Core Functionality and leave this heading as the details for element targeting. This section is buried under Extending Experiences since it's the sort of stuff that almost nobody caries about.

Sources/AppcuesKit/AppcuesKit.docc/AnchoredTooltips.md
```diff
@@ -0,0 +1,51 @@
...
+
+iOS Applications using SwiftUI can also identify `View` elements. A `View` extension `appcuesView(identifier:)` is provided by the Appcues iOS SDK to support this use case. The `identifier` String value must be unique on the screen where an anchored tooltip may be targeted.
+
+The following example shows how this would be added to a SwiftUI `Button` View:
+
+```swift
+Button("Save Profile") {
+    onSaveProfile()
+}
+.appcuesView("btnSaveProfile")
```
mmaatttt, line 40: Should be `.appcuesView(identifier: "btnSaveProfile")`

mmaatttt (approved):
Just fix the one snippet and we're good to go. This is excellent!

Sources/AppcuesKit/AppcuesKit.docc/AnchoredTooltips.md
iujames: ah thank you, fixing

Sources/AppcuesKit/AppcuesKit.docc/AppcuesKit.md
iujames: yep, i like that. updating

PR #426, opened 2023-06-27
Update embed rendering logic

mmaatttt:
1. properly cache a single experience from ExperienceLoader (previously it was only caching on an error, which is wrong because we want to handle the disappear/reappear scenario, and because the `show()` function is used internally by `ExperienceRenderer`, it was replacing multiple cached items with the single one)
2. handle cache clearing on dismissal (there's an open question [here](https://github.com/appcues/appcues-mobile-experience-spec/pull/194#discussion_r1243704479), but I've implemented the approach that I thought made sense)
3. avoid attempting to render multiple experiences when no context is available (special case to avoid trying and failing every embed)
4. add `experience_error` and `experience_recovered` handling

### DIFF
--- a/Sources/AppcuesKit/Presentation/ExperienceRendering/ExperienceRenderer.swift
+++ b/Sources/AppcuesKit/Presentation/ExperienceRendering/ExperienceRenderer.swift
@@ -66,7 +66,6 @@ internal class ExperienceRenderer: ExperienceRendering, StateMachineOwning {
     func processAndShow(qualifiedExperiences: [ExperienceData], reason: ExperienceTrigger) {
         let shouldClearCache = reason == .qualification(reason: .screenView)
         if shouldClearCache {
-            // TODO: error events for embeds that didn't show because there was no Frame
             potentiallyRenderableExperiences = [:]
             stateMachines.cleanup()
         }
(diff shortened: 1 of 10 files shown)


Sources/AppcuesKit/Presentation/ExperienceRendering/ExperienceRenderer.swift
```diff
@@ -92,7 +97,7 @@ internal class ExperienceRenderer: ExperienceRendering, StateMachineOwning {
         }
 
         guard let stateMachine = stateMachines[experience.renderContext] else {
-            potentiallyRenderableExperiences[experience.renderContext] = [experience]
+            analyticsObserver.trackRecoverableError(experience: experience, message: "no render context \(experience.renderContext)")
```
iujames, line 100: Checking understanding - so the addition to cache is not needed here any longer since that is handled in the new L86. Pretty sure that is correct 👍

iujames (approved):
Looks good, lines up with latest chat on the design doc 🚀

Sources/AppcuesKit/Presentation/ExperienceRendering/ExperienceRenderer.swift
mmaatttt: Yep, exactly. This was problematic for the reasons described in pt 1 of the description.

PR #428, opened 2023-07-11
Update example apps with embed examples

mmaatttt:
Updating the SPM and Cocoapods examples with the Embed tab and a sub-page that can be launched from the button. The main page has 4 frames (frame1-frame4). The subpage allows toggling 2 frame (frame1, frame2) and will be useful for playing with the most complex sequences.

New `appcues-example://embed` deep link also added.

|Main tab|Button detail screen|
|-|-|
|Simulator Screenshot - iPhone 14 Pro - 2023-07-11 at 14 26 35|Simulator Screenshot - iPhone 14 Pro - 2023-07-11 at 14 26 38|


iujames (approved):
awesome! only idea would be maybe to have the action be `Toggle "frame1"` instead of `Toggle Frame 1` just so it is very clear what the `frameID` value should be for testing. But that's pretty simple and obvious after you do it once so maybe not needed.

PR #430, opened 2023-07-24
Update session ID when new session started after session timeout

iujames:


mmaatttt (approved):
Thanks for updating the tests too!

PR #431, opened 2023-07-31
Ensure no empty strings are present in selector property values

iujames:
Some customer data showed selectors with `""` values - we should ignore these and not create selectors that will not provide any unique value.

### DIFF
--- a/Sources/AppcuesKit/Presentation/Debugger/ScreenCapturing/UIKitElementTargeting.swift
+++ b/Sources/AppcuesKit/Presentation/Debugger/ScreenCapturing/UIKitElementTargeting.swift
@@ -145,11 +145,11 @@ internal extension UIView {
 
     private func getAppcuesSelector(autoTag: String? = nil) -> UIKitElementSelector? {
         return UIKitElementSelector(
-            appcuesID: (self as? AppcuesTargetView)?.appcuesID,
-            accessibilityIdentifier: accessibilityIdentifier,
-            accessibilityLabel: accessibilityLabel,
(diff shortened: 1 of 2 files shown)


Sources/AppcuesKit/Presentation/Debugger/ScreenCapturing/UIKitElementTargeting.swift
```diff
@@ -145,11 +145,11 @@ internal extension UIView {
...
         return UIKitElementSelector(
-            appcuesID: (self as? AppcuesTargetView)?.appcuesID,
-            accessibilityIdentifier: accessibilityIdentifier,
-            accessibilityLabel: accessibilityLabel,
+            appcuesID: (self as? AppcuesTargetView)?.appcuesID.emptyToNil(),
+            accessibilityIdentifier: accessibilityIdentifier.emptyToNil(),
+            accessibilityLabel: accessibilityLabel.emptyToNil(),
             tag: tag != 0 ? "\(self.tag)" : nil,
-            autoTag: autoTag
+            autoTag: autoTag.emptyToNil()
```
mmaatttt, line 152: This works fine, but I think the right place to do this is in the `UIKitElementSelector.init()`. In case there's ever anything else that's creating a selector object, we'd want the same "no empty strings" rule to apply.

iujames: sure, good point - i'll move the logic there

iujames: updated!

mmaatttt (approved):

PR #435, opened 2023-08-11
Embed enhancements

mmaatttt:
1. For the case where a FrameView(id: A) is registered and then a different FrameView(id: A) is registered. Now We call `reset()` for the first FrameView which removes the appcues content and removes the state machine without triggering any analytics. This essential resets the first FrameView back to how it was before it was registered. This is ideal because the newly registered second frame replaces the first in the state machine directory.
2. `AppcuesFrame` (the SwiftUI implementation) doesn't need the non-zero `preferredContentSize`, and having it created an additional gap in the layout.
3. EDIT: There's also the case where the same FrameView is registered for multiple IDs. Previously that meant the same StateMachineOwning would be in the directory for multiple RenderContexts, causing all kinds of issues. Now if a frame is registered for a new ID, we reset the contents (if there was something showing for ID A, but the ID is now B, it shouldn't show the content for A. This is essentially the `prepareForReuse` scenario). We also enforce uniqueness in the `StateMachineDirectory`, removing the old RenderContext reference before adding the new one.

### DIFF
--- a/Sources/AppcuesKit/Presentation/ExperienceRendering/ExperienceRenderer.swift
+++ b/Sources/AppcuesKit/Presentation/ExperienceRendering/ExperienceRenderer.swift
@@ -33,8 +33,10 @@ internal enum ExperienceRendererError: Error {
 @available(iOS 13.0, *)
 internal class ExperienceRenderer: ExperienceRendering, StateMachineOwning {
 
+    // Conformance to `StateMachineOwning`.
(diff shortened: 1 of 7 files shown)


mmaatttt:
@iujames I added a new commit and pt 3 in the description if you don't mind taking another look.

iujames (approved):
Nice, lots of details I never fully imagined, but nice work chasing them all down.

PR #436, opened 2023-08-14
Fix embed rendering

mmaatttt:
Adjusting the embed trait to be container decorating addresses [an issue](https://appcues.slack.com/archives/C031RLGS4F8/p1692023469654459) where the frame position and size would appear wonkily. This is caused by an early call to the containerVC.view, which triggers viewDidLoad early and means there's a dropped content size update. This change works because `AppcuesWrapperCreatingTrait` gets called early in the trait composer vs `AppcuesContainerDecoratingTrait` which gets called later.

In all my playing around trying to resolve the above issue, I also finally addressed the spots we have the "min size 1" requirement. That requirement was caused by the carousel trait, and so the fix for it should be self contained in the carousel trait rather than having code everywhere (including in the WIP [RN](https://github.com/appcues/appcues-react-native-module/blob/c4842074d4ee8b603a1876339fe1cc2921a79329/ios/AppcuesFrameViewManager.swift#L72-L79) and [Flutter](https://github.com/appcues/appcues-flutter-plugin/blob/77457d00c71ee5d3a4b8e470a1b2b91678a65d4d/ios/Classes/Embeds/AppcuesFrameViewFactory.swift#L119-L126) implementations) to make sure the view has its min size. (The RN and Flutter implementation should be able to do `let size = isHidden ? .zero : preferredContentSize` now.)

### DIFF
--- a/Sources/AppcuesKit/Presentation/Traits/Appcues/AppcuesCarouselTrait.swift
+++ b/Sources/AppcuesKit/Presentation/Traits/Appcues/AppcuesCarouselTrait.swift
@@ -245,11 +245,18 @@ extension AppcuesCarouselTrait {
             super.init(frame: .zero)
 
             addSubview(collectionView)
+
+            let bottomConstraint = collectionView.bottomAnchor.constraint(equalTo: bottomAnchor)
+            bottomConstraint.priority = .init(999)
+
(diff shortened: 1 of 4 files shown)


iujames (approved):
awesome! great to have the centralized fix for the carousel

PR #437, opened 2023-08-16
Update Trait Application Sequence diagram

mmaatttt:
Correcting this for posterity before deleting it


iujames (approved):
Hope to see it revived some day!

PR #439, opened 2023-08-16
Reduce visibility for experience actions and traits

mmaatttt:
Anything that was `public` and shouldn't be anymore, I changed to `internal` even if some of the properties or functions don't explicitly need access control. There were also a few places that were marked `internal` in `public` classes that I've removed the access control identifier for to differentiate from the first case of things that used to be `public`.

Also added a migration guide for 2.x->3.0. The 1.x->2.0 migration guide was a mess since none of the things being referenced with double-backticks exist anymore, so I removed the references for regular code blocks. It's kinda silly to keep the doc at all, but I like it a general best practice.


Sources/AppcuesKit/AppcuesKit.docc/3.0MigrationGuide.md
```diff
@@ -0,0 +1,11 @@
...
+
+## Overview
+
+Updating to this release will not require any code changes for most SDK installations.  Code changes would only be required if your app implemented custom experience traits or actions.
+
+## General Changes
+
+### Removed
+
+- To simplify the public API, `AppcuesExperienceAction` and `AppcuesExperienceTrait` related classes and functions no longer have public visibility. There no current use cases for extensibility that require these to be public.
```
iujames, line 11: > There no current use cases

There **are** no current use cases

I see this is an error from the Android side, I'll fix that!

iujames (approved):

Sources/AppcuesKit/AppcuesKit.docc/3.0MigrationGuide.md
mmaatttt: Fixed this

PR #440, opened 2023-08-16
Session handling updates

iujames:
Similar to what was done in https://github.com/appcues/appcues-android-sdk/pull/452, but thankfully less invasive here. Aligning on Appcues session definition per [doc](https://www.notion.so/appcues/The-definition-of-a-session-at-Appcues-bbaa1c717e9146fcadddaa4a96daea5c?pvs=4)

high level: remove session events for suspend/resume/reset - only track session_start. Update the timeout default to 300 sec (was 1800), and check for timeout after any period of analytics inactivity, regardless of app background/foreground. No longer track background/foreground at all effectively.

One update to the dependency graph - `AnalyticsPublishing` now uses `SessionMonitoring`, rather than the other way around. The red arrow here thus had its direction flipped ([link to final result](https://whimsical.com/ios-sdk-dependency-graph-UHHdjHhVSixeGWttw3HZzH))

Screenshot 2023-08-16 at 4 28 22 PM

The bulk of the important logic here now is in `AnalyticsPublishing`, where every time a new tracking update comes in, it checks to see if we either don't have a session yet, or it has expired. In either case, it attempts to start a new one. This may fail, if no user. If it succeeds - a `session_started` event is immediately tracked and the new session begins.

`SessionMonitor` is simplified, and we no longer listen to foregrounding at all, and only listen to backgrounding to attempt flush analytics if possible. Some related tests become invalid now, and are removed, others updated for new changes.

### DIFF
--- a/Sources/AppcuesKit/Appcues.swift
+++ b/Sources/AppcuesKit/Appcues.swift
@@ -162,7 +162,6 @@ public class Appcues: NSObject {
     /// Clears out the current user in this session. Can be used when the user logs out of your application.
     @objc
     public func reset() {
-        // call this first to close final analytics on the session
         sessionMonitor.reset()
 
         storage.userID = ""
@@ -337,9 +336,6 @@ public class Appcues: NSObject {
         // register the auto property decorator
         let autoPropDecorator = container.resolve(AutoPropertyDecorator.self)
         analyticsPublisher.register(decorator: autoPropDecorator)
-
-        // start session monitoring
-        sessionMonitor.start()
     }
 
     private func identify(isAnonymous: Bool, userID: String, properties: [String: Any]? = nil) {
@@ -355,7 +351,7 @@ public class Appcues: NSObject {
         storage.userSignature = properties?.removeValue(forKey: "appcues:user_id_signature") as? String
         if userChanged {
             // when the identified user changes from last known value, we must start a new session
-            sessionMonitor.start()
+            sessionMonitor.reset()
 
             // and clear any stored group information - will have to be reset as needed
             storage.groupID = nil
--- a/Sources/AppcuesKit/Data/Analytics/SessionMonitor.swift
+++ b/Sources/AppcuesKit/Data/Analytics/SessionMonitor.swift
@@ -10,15 +10,15 @@ import Foundation
 import UIKit
 
 internal protocol SessionMonitoring: AnyObject {
-    func start()
+    var isSessionExpired: Bool { get }
+
+    func start() -> Bool
+    func updateLastActivity()
     func reset()
 }
 
 internal enum SessionEvents: String, CaseIterable {
     case sessionStarted = "appcues:session_started"
-    case sessionSuspended = "appcues:session_suspended"
-    case sessionResumed = "appcues:session_resumed"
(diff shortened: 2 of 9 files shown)


Sources/AppcuesKit/Appcues.swift
```diff
@@ -337,9 +336,6 @@ public class Appcues: NSObject {
         // register the auto property decorator
         let autoPropDecorator = container.resolve(AutoPropertyDecorator.self)
         analyticsPublisher.register(decorator: autoPropDecorator)
-
-        // start session monitoring
-        sessionMonitor.start()
```
iujames, line 342: no longer needed as sessions are controlled by `AnalyticsPublisher` on demand

Sources/AppcuesKit/Appcues.swift
```diff
@@ -355,7 +351,7 @@ public class Appcues: NSObject {
         storage.userSignature = properties?.removeValue(forKey: "appcues:user_id_signature") as? String
         if userChanged {
             // when the identified user changes from last known value, we must start a new session
-            sessionMonitor.start()
+            sessionMonitor.reset()
```
iujames, line 354: when the user changes, now we clear the session, and the subsequent `identify()` will start a new one

Sources/AppcuesKit/Presentation/Debugger/UIDebugger.swift
```diff
@@ -132,6 +132,8 @@ internal class UIDebugger: UIDebugging {
     @objc
     private func appcuesReset(notification: Notification) {
         self.viewModel.reset()
+        self.viewModel.currentUserID = self.storage.userID
+        self.viewModel.isAnonymous = self.storage.isAnonymous
```
iujames, line 136: detail here - clear out the debugger current user on a reset, rather than rely on reset event

mmaatttt (approved):
Good stuff!

Sources/AppcuesKit/Data/Analytics/SessionMonitor.swift
```diff
@@ -56,42 +54,32 @@ internal class SessionMonitor: SessionMonitoring {
     }
 
     // called on (A) sdk init (B) user identity change
```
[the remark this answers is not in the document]
iujames: yep good catch - will update

Sources/AppcuesKit/Data/Analytics/SessionMonitor.swift
```diff
@@ -56,42 +54,32 @@ internal class SessionMonitor: SessionMonitoring {
...
 
     // called on reset(), user sign-out
     func reset() {
-        // this is interactive: true since a reset should flush to network immediately (with previous user ID)
-        // and the next session start will be sent in a new request, with the new user ID
-        publisher.track(SessionEvents.sessionReset, properties: nil, interactive: true)
-        appcues?.sessionID = nil
-    }
+        // ensure any pending in-memory analytics get processed asap
+        tracker.flush()
```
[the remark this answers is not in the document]
iujames: yep thats correct - I think technically it might work still if things just went out on the queue, but wanted to be safe since the user ID and session ID are getting cleared out right after this.

PR #441, opened 2023-08-21
Update toasts and experience previews

mmaatttt:
The user-facing goal here is to make previewing embed experiences easier. This is accomplished two ways:

1. Maintain different cache rules for experience previews so that the preview data will persist across screen view events. This will allow a user who's triggered a preview to navigate to the appropriate screen in their app to view the preview.
2. Add a toast message explaining that a preview is pending (instead of the app opening with nothing happening).

To accomplish 2, I modified our existing toast system for screen capture. Previous it was very tightly coupled to a few specific messages in the context of the debugger flow (ie a lot of code in `DebugView` and `DebugViewController`). I've refactored it substantially (mostly just moving code out of those files) to work with a new `DebugToast` model and a `ToastUIWindow`. Now the only direct connection to the debugger is that it's responsible for creating the `ToastUIWindow`. In the future, if we want to explore toasts as a new pattern, the work I've done here will hopefully give us a solid foundation.

I've also added error handling for other experience preview errors. I think it's much nicer to get a toast explaining that the preview failed rather than just having the app open and nothing change.

|Embed preview|Bad experience ID in deep link|
|-|-|
|Simulator Screenshot - iPhone 14 Pro - 2023-08-21 at 10 32 55|Simulator Screenshot - iPhone 14 Pro - 2023-08-21 at 10 33 03|


Sources/AppcuesKit/Presentation/DeepLinkHandler.swift
```diff
@@ -131,6 +131,27 @@ internal class DeepLinkHandler: DeepLinkHandling {
...
+            message = "Experience not found."
+        case is NetworkingError:
+            message = "Error loading mobile flow preview."
+        default:
+            message = "Mobile flow preview failed."
+        }
+
+        let toast = DebugToast(message: .custom(text: message), style: .failure)
+        container?.resolve(UIDebugging.self).showToast(toast)
+    }
```
iujames, line 153: nice, awesome that we already had the completion plumbing in place! this fit in nicely. So a failed tooltip for a missing target would just be `Mobile flow preview failed` right? I think thats good, nicely delineates the possible reasons something might not show.

@andretortolano I think we'll have to change `ExperienceRenderer.show()` to return a `ResultOf` instead of a `Boolean` to do something like this and get the actual reasons for failure.

iujames (approved):
Great work extracting and generalizing the toast view, that's nice to have!

Sources/AppcuesKit/Presentation/DeepLinkHandler.swift
mmaatttt: Correct. I can add a case for `ExperienceStateMachine.ExperienceError` and show the experience error if we want.

(I can also add support for >2 lines)

Simulator Screenshot - iPhone 14 Pro - 2023-08-21 at 13 36 13

PR #447, opened 2023-08-24
Embed polishing

mmaatttt:
3 little things:
1.  Transition frame state machine to idling instead of setting to nil
2. Fixing the fade animation
3. Include frameID in debugger experience list


Sources/AppcuesKit/Presentation/ExperienceRendering/ExperienceRenderer.swift
```diff
@@ -314,7 +314,7 @@ internal class ExperienceRenderer: ExperienceRendering, StateMachineOwning {
 
     /// Reset only the owned `Self.stateMachine` instance in conformance to `StateMachineOwning`.
     func reset() {
-        stateMachine?.removeObserver(analyticsObserver)
+        stateMachine?.removeAnalyticsObserver()
```
mmaatttt, line 317: Changed this because the AppcuesFrameView doesn't have a reference to the instance to use the previous `removeObserver`.

iujames (approved):

PR #450, opened 2023-09-05
Update text init to avoid inadvertant localization

mmaatttt:
Just in case an app has `""` as a localization key.

### DIFF
--- a/Sources/AppcuesKit/Presentation/UI/Components/AppcuesText.swift
+++ b/Sources/AppcuesKit/Presentation/UI/Components/AppcuesText.swift
@@ -32,7 +32,7 @@ internal struct AppcuesText: View {
 @available(iOS 13.0, *)
 extension Text {
     init(textModel: ExperienceComponent.TextModel, skipColor: Bool = false, scaled: Bool = false) {
-        self.init("")
+        self.init(verbatim: "")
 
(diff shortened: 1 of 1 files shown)


Sources/AppcuesKit/Presentation/UI/Components/AppcuesText.swift
```diff
@@ -32,7 +32,7 @@ internal struct AppcuesText: View {
 @available(iOS 13.0, *)
 extension Text {
     init(textModel: ExperienceComponent.TextModel, skipColor: Bool = false, scaled: Bool = false) {
-        self.init("")
+        self.init(verbatim: "")
```
iujames, line 35: should we do it on L40 as well?

mmaatttt: I don't think we need to since a stored value isn't treated the same as a string literal:

>By default, SwiftUI assumes that you don’t want to localize stored strings, but if you do, you can first create a localized string key from the value, and initialize the text view with that. Using a key as input triggers the `init(_:tableName:bundle:comment:)` method instead.

- https://developer.apple.com/documentation/swiftui/text/init(_:)-9d1g4

We can though if we want to be super clear.

iujames (approved):

PR #453, opened 2023-09-13
Add session_id to the root of activity/qualify API request body

iujames:
This approach has the `AnalyticsTracker` keep a reference to `Appcues` so that it can grab the session ID at the moment it translates a `TrackingUpdate` to `Activity`. There are a few ways we could do this. I think this is the smallest change, so starting here for any feedback.

Other possibilities:
1. Have the `TrackingUpdate` contain the session ID within. This could be done. There are six different classes that can create a `TrackingUpdate` currently, and most of those spots would need updated with handling to check for the session ID and pass through - somewhere along the line still checking for nil.
2. Update the `AnalyticsSubscriber.track(update)` function to also pass the session ID here, as second param - this would allow `AnalyticsPublisher`, which already has the `Appcues` reference and already checks for session, to pass it through to the `AnalyticsTracker` which could then use it to create the `Activity`. It felt a little odd to change this interface for this prop, but its internal and only used in two places I think, so eh. It may be the most appropriate spot to handle the nil check (already doing it)
3. others?

I think these are all pretty much equally viable, so happy to go a different route if there is a preference.

### DIFF
--- a/Sources/AppcuesKit/Data/Analytics/AnalyticsTracker.swift
+++ b/Sources/AppcuesKit/Data/Analytics/AnalyticsTracker.swift
@@ -16,6 +16,8 @@ internal class AnalyticsTracker: AnalyticsTracking, AnalyticsSubscribing {
 
     private static let flushAfterSeconds: Double = 10
 
+    private weak var appcues: Appcues?
+
     private let storage: DataStoring
     private let config: Appcues.Config
     private let activityProcessor: ActivityProcessing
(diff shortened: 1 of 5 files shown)


mmaatttt (approved):
I think this is a good approach.

Generally speaking, it would seem more conceptually correct if `SessionMonitor` owned the `sessionID` instead of `Appcues`. But then `AutoPropertyDecorator` and `AnalyticsTracker` would need to to resolve `SessionMonitor` which causes circular dependencies and other changes, so I don't think it's worthwhile to pursue.

PR #455, opened 2023-09-20
Localization enhancements

mmaatttt:
1. Fix the debugger to always be LTR since it's always in English
2. Update the `_lastBrowserLanguage` auto property to use the bundle value
3. Add `localeId` and `localeName` to experience analytics if the `context` values are provided

### DIFF
--- a/Sources/AppcuesKit/Data/Models/Experience.swift
+++ b/Sources/AppcuesKit/Data/Models/Experience.swift
@@ -45,6 +45,7 @@ internal struct FailedExperience: Decodable {
             name: name ?? "",
             type: type ?? "",
             publishedAt: publishedAt,
+            context: nil,
             traits: [],
             steps: [],
             redirectURL: nil,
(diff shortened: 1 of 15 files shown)


Sources/AppcuesKit/Data/Models/Experience.swift
```diff
@@ -45,6 +45,7 @@ internal struct FailedExperience: Decodable {
             name: name ?? "",
             type: type ?? "",
             publishedAt: publishedAt,
+            context: nil,
```
iujames, line 48: on Android side, I added the optional `context` on the failed version as well, since it may lend some insight into the experience error that gets generated - nice to know which localized version it came from, if available. I think harmless to try, if optional on the deserialization.

iujames (approved):

Sources/AppcuesKit/Data/Models/Experience.swift
mmaatttt: Good idea. Updated in 12c7b70.

PR #456, opened 2023-09-26
Add test for analytics queuing and response handling

mmaatttt:
To make analytics queue processing testable, I moved the `private static flushAfterSeconds` to the `Config` object so I can modify it in the test. The test still takes 3s which is long for a unit test, but a lot better than waiting 11s.

Also added a couple tests that verify `AnalyticsTracker` calls `ExperienceRenderer` the way we think it does.

### DIFF
--- a/Tests/AppcuesKitTests/Analytics/AnalyticsTrackerTests.swift
+++ b/Tests/AppcuesKitTests/Analytics/AnalyticsTrackerTests.swift
@@ -192,6 +192,95 @@ class AnalyticsTrackerTests: XCTestCase {
         // Assert
         waitForExpectations(timeout: 1)
     }
+
+    func testQueuedEvents() throws {
+        // Arrange
+        appcues.config.flushAfterDuration = 2
+
+        let minTimeExpectation = expectation(description: "Valid request")
+        let onRequestExpectation = expectation(description: "Waited the flush duration")
+        let expectedEvents = [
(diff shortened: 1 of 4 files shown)


Tests/AppcuesKitTests/Analytics/AnalyticsTrackerTests.swift
```diff
@@ -192,6 +192,97 @@ class AnalyticsTrackerTests: XCTestCase {
...
+        appcues.config.flushAfterDuration = 2
+
+        let minTimeExpectation = expectation(description: "Valid request")
+        let onRequestExpectation = expectation(description: "Waited the flush duration")
+        let expectedEvents = [
+            Event(name: "appcues:v2:step_interaction", attributes: ["my_key": "my_value", "another_key": 33]),
+            Event(name: "appcues:v2:step_interaction", attributes: ["another_key": 100])
+        ]
+
+        // In the case of a profile update, verify the update is sent to the activity processor as
```
iujames, line 207: > In the case of a profile update

maybe a copy/paste?

iujames (approved):

Tests/AppcuesKitTests/Analytics/AnalyticsTrackerTests.swift
mmaatttt: yep 😅

PR #457, opened 2023-09-27
Add support for registering fonts from a bundle path

mmaatttt:
The `Appcues.Config` function registers the fonts and saves the path for the debugger to reference.

I refactored the debugger font view a bit since the view shouldn't be responsible for determining the font list. So now the view model does.

### DIFF
--- a/Sources/AppcuesKit/Appcues+Config.swift
+++ b/Sources/AppcuesKit/Appcues+Config.swift
@@ -48,6 +48,8 @@ public extension Appcues {
 
         var enableTextScaling = false
 
+        var bundleFontsPath: String?
+
         /// Create an Appcues SDK configuration
         /// - Parameter accountID: Appcues Account ID - a string containing an integer, copied from the Account settings page in Studio.
         /// - Parameter applicationID: Appcues Application ID - a string containing a UUID,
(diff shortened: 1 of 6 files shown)


Sources/AppcuesKit/Presentation/Debugger/Panel/DebugViewModel.swift
```diff
@@ -86,18 +87,58 @@ internal class DebugViewModel: ObservableObject {
...
         self.userIdentified = !currentUserID.isEmpty
+        self.bundleFontsPath = bundleFontsPath
 
         connectedStatus.action = Action(symbolName: "arrow.triangle.2.circlepath") { [weak self] in self?.ping() }
         deepLinkStatus.action = Action(symbolName: "arrow.triangle.2.circlepath") { [weak self] in self?.verifyDeepLink() }
     }
 
+    func fonts() -> [(title: String, names: [String])] {
+        let uiAppFontPaths = (Bundle.main.infoDictionary?["UIAppFonts"] as? [String] ?? []).compactMap { Bundle.main.url(forResource: $0, withExtension: nil) }
+        let customBundleFontPaths = Bundle.main.urls(forResourcesWithExtension: "ttf", subdirectory: bundleFontsPath) ?? []
```
mmaatttt, line 112: This is the important link to read from the bundle fonts. The rest of this is just refactored from `DebugFontUI`.

Sources/AppcuesKit/Appcues+Config.swift
```diff
@@ -194,5 +196,50 @@ public extension Appcues {
...
+                    logger.info(
+                        "No fonts found in main bundle in subdirectory %{public}s",
+                        path
+                    )
+                }
+
+                fontResources.forEach { fontURL in
+                    var error: Unmanaged<CFError>?
+
+                    if !CTFontManagerRegisterFontsForURL(fontURL as CFURL, .process, &error) {
```
mmaatttt, line 227: There are different scope options: https://developer.apple.com/documentation/coretext/ctfontmanagerscope

`.process` seems right from the descriptions above, but would appreciate a sanity check.

iujames: yeah, seems like it. I don't really understand what "current use session" would mean and how that would differ from process though.

PR #458, opened 2023-10-02
Xcode 15 unit test updates

iujames:
[sc-57722]

Xcode 15 introduced a couple of small changes that we need to revise our unit tests to match.

1. The generated JSON decoding error messages changed slightly, and this affects the comparison we're doing in the tests around decoding malformed experiences.
2. It appears we cannot directly compare two `Data` objects for equivalent contents any longer - so we need to actually decode and inspect the `Data` being POST-ed in our network layer during the ActivityProcessorTest


iujames:
this will need to wait, as CircleCI has issues with Xcode 15, causing the job to consistently time out.
https://discuss.circleci.com/t/severe-performance-problems-with-xcode-15/49205

PR #461, opened 2023-10-17
Debugger refactor

mmaatttt:
The debugger is undoubtedly (by far!) in the poorest shape of any part of the SDK codebase and is almost entirely untested. It's not a critical piece of the SDK, so that's not a huge problem, but with stuff like screen capture being increasingly important, it'd be worthwhile trying to make it better. This PR does some significant refactoring and adds a bunch of new tests.

## Changes
1. Break up `DebugViewModel`. This >500 line file has a massive amount of stuff in it.
  1. I've moved models like `StatusItem` and `LoggedEvent` to their own files.
  2. I've created standalone classes `APIVerifier` and `DeepLinkVerifier` that encapsulate that functionality.
2. `DebugUI` now has views for each of the Status items. This better separates concerns and will make testing these in the future easier (all the tests in this PR are just related to the logic—nothing about the presentation yet).
3. I've created `ScreenCapturer which encapsulate the screen capture logic instead of it being bolted on to `UIDebugger`. It uses the new `protocol ScreenCaptureUI` to communicate back to `UIDebugger`.
4. Moved `LifecycleEvent.EventProperties` to the debugger directories as `StructuredLifecycleProperties` since the only place it's used is in the debugger.

## Benefits
1. Testability! There's now tests for all this stuff.
2. The DebugViewController and friends are tied the the lifecycle of the debug view and not the UIDebugger class. So when the debugger is dismissed, all that stuff is deinited instead of sticking around forever after it's shown once.
3. Simpler data flow. The events that come in via AnalyticsSubscribing are now piped through a single combine Publisher that gets used to show the fleeting log items, update the event list in the debugger, and calculate the active experiences section of the debugger.

## How to review this PR
Looking at the final diff will make it quite challenging to follow where everything moved to. Reviewing each of the 3 "♻️ Refactor..." commits might be simpler. Note that some of the things in ac6c4f0 were subsequently changed further in 118543b (like where the `APIVerifier` and `DeepLinkVerifier` are inited).

Some of the tests are complicated. For example, the ScreenCapturer still does the 4 changed network requests, so to properly mock/test the last step, we need to mock/test all 3 before it. Also testing the Combine pipeline stuff is a bit awkward since you need to set an expectation to fulfill once you see the values you want coming through the pipeline.

### DIFF
--- a/Sources/AppcuesKit/Presentation/Debugger/Panel/DebugFontUI.swift
+++ b/Sources/AppcuesKit/Presentation/Debugger/Panel/DebugFontUI.swift
@@ -12,14 +12,42 @@ import SwiftUI
 @available(iOS 13.0, *)
 internal enum DebugFontUI {
     struct FontListView: View {
-        let sections: [(title: String, names: [String])]
+        static let sections: [(title: String, names: [String])] = {
+            let familyNames: [String] = (Bundle.main.infoDictionary?["UIAppFonts"] as? [String] ?? [])
+                .compactMap { resourceName in
+                    if let url = Bundle.main.url(forResource: resourceName, withExtension: nil),
+                       let fontData = try? Data(contentsOf: url),
+                       let fontDataProvider = CGDataProvider(data: fontData as CFData),
+                       let font = CGFont(fontDataProvider),
+                       let name = font.postScriptName {
+                        let ctfont = CTFontCreateWithName(name, 17, nil)
+                        return CTFontCopyFamilyName(ctfont) as String
+                    } else {
+                        return nil
+                    }
+                }
+
+            let appFonts = Set(familyNames).sorted().flatMap { UIFont.fontNames(forFamilyName: $0) }
+            let systemFonts = Font.Design.allCases.flatMap { design in
+                Font.Weight.allCases.map { weight in
+                    "System \(design.description) \(weight.description)"
+                }
+            }
+            let allFonts = UIFont.familyNames.flatMap { UIFont.fontNames(forFamilyName: $0) }
+
+            return [
+                ("App-Specific Fonts", appFonts),
+                ("System Fonts", systemFonts),
+                ("All Fonts", allFonts)
+            ]
+        }()
 
         private var filteredSections: [(title: String, names: [String])] {
             let query = searchText.lowercased()
 
-            guard !query.isEmpty else { return sections }
+            guard !query.isEmpty else { return FontListView.sections }
 
-            return sections.compactMap { title, names in
+            return FontListView.sections.compactMap { title, names in
                 let filteredNames = names.filter { $0.lowercased().contains(query) }
                 if !filteredNames.isEmpty {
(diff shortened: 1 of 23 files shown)


Sources/AppcuesKit/Presentation/Debugger/Panel/DebugFontUI.swift
```diff
@@ -12,14 +12,42 @@ import SwiftUI
 @available(iOS 13.0, *)
 internal enum DebugFontUI {
     struct FontListView: View {
-        let sections: [(title: String, names: [String])]
+        static let sections: [(title: String, names: [String])] = {
+            let familyNames: [String] = (Bundle.main.infoDictionary?["UIAppFonts"] as? [String] ?? [])
```
iujames, line 16: noting that [previously](https://github.com/appcues/appcues-ios-sdk/pull/457) there was an attempt to move the font list calculation out of the view layer - this seems to reverse back. Any downside to this?

Tests/AppcuesKitTests/Debugger/APIVerifierTests.swift
```diff
@@ -0,0 +1,81 @@
...
+        networking.onGet = { endpoint, _, completion in
+            completion(.failure(NetworkingError.nonSuccessfulStatusCode(500)))
+        }
+
+        // Act
+        apiVerifier.verifyAPI()
+
+        // Assert
+        wait(for: [expectation], timeout: 1)
+        XCTAssertEqual(values.count, 2, "a pending value and a success one")
```
iujames, line 77: `and a unverified one`

Tests/AppcuesKitTests/Debugger/DeepLinkVerifierTests.swift
```diff
@@ -0,0 +1,180 @@
...
+
+
+        deepLinkVerifier.mockURLTypes = [["CFBundleURLSchemes": ["appcues-<unsafe>"]]]
+
+        // Act
+        deepLinkVerifier.verifyDeepLink()
+
+        // Assert
+        wait(for: [expectation], timeout: 1)
+        XCTAssertEqual(values.count, 2, "a pending value and a success one")
```
iujames, line 77: `and an unverified one` (also L105, L135, L165)

iujames (approved):
Looks great! slick Combine usage and great code structure clean up and improved testing

Sources/AppcuesKit/Presentation/Debugger/Panel/DebugFontUI.swift
mmaatttt: The previous change ended up being very inefficient since the list was recalculated every time the debug view refreshed. A static property here means the list gets calculated once, when the view appears. I could move the static property somewhere else...maybe an extension on `Font` would make sense.

mmaatttt:
Actual tests passed. The codecov updater failed to download.

PR #463, opened 2023-10-19
Add identify to example embed options

mmaatttt:
Test app update that lets us identify a user from the embed test harness. Merging this before the re-render fix so we can verify the bug and fix in the TestFlight app.


iujames (approved):

mmaatttt:
Codecov uploader failed again 🙃

PR #466, opened 2023-10-24
Remove _userAgent auto property

mmaatttt:
Removing this auto property on iOS for performance reasons: the user agent value was determined at SDK init and must be on the main thread because it uses a UIKit view.

It is not believed to be a useful property in mobile for most scenarios, and it can be sent as a custom property if still desired for targeting content.

Sibling of https://github.com/appcues/appcues-android-sdk/pull/520


iujames (approved):
Just a reminder - you are probably already on it - but these [analytics snapshots](https://github.com/appcues/appcues-mobile-experience-spec/tree/main/verification-suites/ios/UITests/__Snapshots__/ExperienceTests) (that don't run in CI) will need updated to match

PR #467, opened 2023-10-24
Add support for AppcuesUniversalLinkHostAllowList Info.plist allow list

mmaatttt:
React Native apps that [support universal links](https://reactnative.dev/docs/linking#enabling-deep-links) always [return true](https://github.com/facebook/react-native/blob/4ed3260266bd1bad0403032179ca426aa7016c68/packages/react-native/Libraries/LinkingIOS/RCTLinkingManager.mm#L81) from `UIApplicationDelegate.application(_:continue:restorationHandler:)`. This means that unless these apps specifically support opening arbitrary links in their universal link handling (which is unlikely), experiences using the link action won't actually open the link because the SDK is told that the app handled the link when the app didn't actually.

This PR adds an easy configuration option (setting the Info.plist is simpler than an `Appcues.Config` value which would need corresponding changes to the React Native module) to provide an allow list of URL hosts that actually are intended for use with universal links.

The implementation is pretty straightforward—I added the Info.plist getter to `URLOpening` so it can be mocked easily in tests.

Updated the docs here, but I also intend to add a new universal link doc to the React Native repo.

### DIFF
--- a/Sources/AppcuesKit/AppcuesKit.docc/UniversalLinking.md
+++ b/Sources/AppcuesKit/AppcuesKit.docc/UniversalLinking.md
@@ -1,10 +1,10 @@
 # Configuring Handling of Universal Links
 
-The Appcues iOS SDK supports universal links as an option deep linking to screens within your app.
+The Appcues iOS SDK supports universal links as an option for deep linking to screens within your app.
 
 ## Overview
 
-You must implement `UIApplicationDelegate.application(_:continue:restorationHandler:)` to handle your universal links. The Appcues iOS SDK will call this method for every `https` link it tries to open to allow your app a chance to handle it internally instead of opening the link in a browser.
+You must implement `UIApplicationDelegate.application(_:continue:restorationHandler:)` to handle your universal links. By default, the Appcues iOS SDK will call this method for every `https` link it tries to open to allow your app a chance to handle it internally instead of opening the link in a browser. Optionally, you may provide an allow list of URL hosts—refer to the Domain Allow List section below.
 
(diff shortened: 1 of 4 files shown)


Sources/AppcuesKit/AppcuesKit.docc/UniversalLinking.md
```diff
@@ -35,3 +35,27 @@ func application(_ application: UIApplication,
     // TODO: Parse the `incomingURL` here and return true if the link has been handled, false otherwise.
 }
 ```
+
+## Implementing a Domain Allow List 
+
+You may define a custom key, `AppcuesUniversalLinkHostAllowList`, in your _Info.plist_ file with an array of (string) host values to control which URLs are passed to `UIApplicationDelegate.application(_:continue:restorationHandler:)`. A URL whose host matches any of the provided host values will be treated as a potential universal link.
+
+An empty array will prevent all links from be treated as universal links and is functionally the same as calling ``Appcues/Config/enableUniversalLinks(_:)`` with a `false` argument.
```
iujames, line 43: nice, was just thinking that merging that config in some fashion in the plist also would be nice.

iujames (approved):

Sources/AppcuesKit/AppcuesKit.docc/UniversalLinking.md
mmaatttt: In the capacitor plugin we're disabling it by default which wouldn't really work with the plist approach.

https://github.com/appcues/appcues-capacitor-plugin/blob/d011729c86b1de005b5418bd66d2cfe81aa6143e/ios/Plugin/AppcuesPlugin.swift#L43-L46

PR #468, opened 2023-10-25
Remove assertionFailures

mmaatttt:
To be as safe as possible, remove places where `assertionFailure` would cause a crash in `-Onone` debug builds (https://developer.apple.com/documentation/swift/assertionfailure(_:file:line:)).

To not lose the message when encoding properties, I'm passing the logging instance through to the encoder from the `Activity` and `Event` models. I'm not the biggest fan of doing this, but I think it's worthwhile (and the parameter is optional).

I have a subsequent PR lined op that makes a generic `Logging` protocol that will let us intercept the logs to show in the debugger and test the log messages that get generated.

### DIFF
--- a/Sources/AppcuesKit/Presentation/Actions/ActionRegistry.swift
+++ b/Sources/AppcuesKit/Presentation/Actions/ActionRegistry.swift
@@ -35,17 +35,12 @@ internal class ActionRegistry {
         register(action: AppcuesRequestReviewAction.self)
     }
 
-    func register(action: AppcuesExperienceAction.Type) {
-        guard actions[action.type] == nil else {
-            #if DEBUG
-            if ProcessInfo.processInfo.environment["XCTestBundlePath"] == nil {
-                assertionFailure("Action of type \(action.type) is already registered.")
-            }
-            #endif
-            return
-        }
+    @discardableResult
+    func register(action: AppcuesExperienceAction.Type) -> Bool {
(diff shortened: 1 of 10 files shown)


Sources/AppcuesKit/Presentation/Actions/ActionRegistry.swift
```diff
@@ -35,17 +35,12 @@ internal class ActionRegistry {
...
-            #if DEBUG
-            if ProcessInfo.processInfo.environment["XCTestBundlePath"] == nil {
-                assertionFailure("Action of type \(action.type) is already registered.")
-            }
-            #endif
-            return
-        }
+    @discardableResult
+    func register(action: AppcuesExperienceAction.Type) -> Bool {
+        guard actions[action.type] == nil else { return false }
```
iujames, line 40: not critical really, now that it is not public, but we're losing the reason why `false` is returned I guess. Maybe just a comment in the api doc in Appcues.swift could say "will not be registered if the given action type is already registered".

iujames (approved):

Sources/AppcuesKit/Presentation/Actions/ActionRegistry.swift
mmaatttt: Good point. I can log an error here easily enough so I'll add that back in along with the comment.

PR #469, opened 2023-10-26
Add log viewing to debugger

mmaatttt:
I used %s for the original logs without much thought (copied from some blog post IIRC), and it worked, but for mapping to an actual string, %@ is the way to go, so fixed that ([format string reference](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/Strings/Articles/formatSpecifiers.html)). We have to strip out the `{public}` and `{private}` OSLog specifiers, but then we can get the log message as a regular String to show in the debugger.

Then the new `Logging` protocol has conformees in `OSLog` and the new `DebugLog` so we can transparently switch between them. `DebugLog` even optionally takes another `Logging` instance so we can continue to send the `OSLog` messages in addition to capturing them in the debugger.

The debug logger only starts when the debugger does (and turns off when the floating button is dismissed) for performance reasons.

For an added cherry on top, we can uses `DebugLogger` in tests to verify the log messages being generated by some failure conditions.

Potentially in the future we could add log filtering and searching, and fancy stuff like that, but I'm not going to worry about it now.

Note that iOS 14+ has a much nicer logging format that doesn't use StaticString, but we can't use it.

|Menu item|List|Detail|
|-|-|-|
|Simulator Screenshot - iPhone 14 Pro - 2023-10-26 at 10 35 06|Simulator Screenshot - iPhone 14 Pro - 2023-10-26 at 10 35 11|Simulator Screenshot - iPhone 14 Pro - 2023-10-26 at 10 35 13|

### DIFF
--- a/Sources/AppcuesKit/Appcues+Config.swift
+++ b/Sources/AppcuesKit/Appcues+Config.swift
@@ -27,7 +27,7 @@ public extension Appcues {
 
         var urlSession: URLSession = NetworkClient.defaultURLSession
 
-        var logger: OSLog = .disabled
+        var logger: Logging = OSLog.disabled
 
         var anonymousIDFactory: () -> String = {
             UIDevice.identifier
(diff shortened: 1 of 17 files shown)


iujames (approved):
very slick! the wrapping of the previous logger, then restoring it on exit is really nice

PR #470, opened 2023-11-06
Add source to activity requests and  _lastSeenAt group auto prop

mmaatttt:
Also [sc-58902]

```diff
{
+	"source": "mobile",
	"account_id": "<account-id>",
	"group_id": "my-group",
	"request_id": "fc592fd8-9210-463b-a73d-c4b19dcfe454",
	"session_id": "1b554130-1d90-4969-8a1d-56e7a877ef11",
	"user_id": "default-00000",
	"group_update": {
+		"_lastSeenAt": 1699296009124
	}
}
```

### DIFF
--- a/Sources/AppcuesKit/Data/Analytics/AutoPropertyDecorator.swift
+++ b/Sources/AppcuesKit/Data/Analytics/AutoPropertyDecorator.swift
@@ -41,6 +41,7 @@ internal class AutoPropertyDecorator: AnalyticsDecorating {
     func decorate(_ tracking: TrackingUpdate) -> TrackingUpdate {
 
         var context = self.contextProperties
+        var decorated = tracking
 
         switch tracking.type {
         case let .screen(title):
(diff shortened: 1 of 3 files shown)


Sources/AppcuesKit/Data/Analytics/AutoPropertyDecorator.swift
```diff
@@ -55,14 +56,15 @@ internal class AutoPropertyDecorator: AnalyticsDecorating {
             currentScreen = nil
             previousScreen = nil
         case .group:
-            // group updates do not have auto props, don't manipulate anything in the properties
-            return tracking
+            // group updates only have this single auto prop, so add that and return early
+            decorated.properties = (decorated.properties ?? [:]).merging([
+                "_lastSeenAt": Date()
+            ])
```
iujames, line 62: double checking - if clearing the group id (setting to nil) does this create properties with a value for the nil group, when there otherwise would not have been a `group_update` in the payload? I suspect we should not send any properties in that case.

mmaatttt: Good catch, I added a case for `nil` group ids

PR #471, opened 2023-11-09
Batch group updates with user identification when called right after identify

iujames:
This is the iOS equivalent of the Android update in https://github.com/appcues/appcues-android-sdk/pull/535 (and similar in concept to the web update in https://github.com/appcues/javascript-sdk/pull/613). Stealing that Android PR description:

* on `identify(id, props)` or the `appcues:session_started` event - a 50ms timer is started to allow analytics that occur immediately after, to be batched together into the same activity payload to the API.
* the expected use case would be a `group(id, props)` called immediately after `identify(id, props)` resulting in a combined payload with the new user and the current group - this prevents any chance of state group information being used during the resulting qualification on the server for the new user update or session start qualification
* all other analytics are unaffected, unless they also happen within that 50ms window (in which case they would go in this immediate batch as well)

This is an example of what this result would look like, for a new user landing on a login page, signing in, and then triggering the resulting user profile update, the session_started event, and an immediate group update to identify that user's current group and props.
Screenshot 2023-11-09 at 3 28 59 PM

This is implemented by adding a new priority queue to our `AnalyticsTracker` - the 50ms queue. We still have the 10 sec background batching queue for flow events as well. If a new priority update comes in (identify or session_start) it will flush anything in the queue existing, and start this new 50ms queue to gather up all immediately subsequent updates.

Couple of edge cases:
* There is already a 50ms queue going for another user - could happen if you made multiple `identify(id)` calls for different user ids, in rapid succession. This will result in the existing priority queue being immediately flushed, stopping any existing timer, before processing the new one.
* There is a 50ms queue running while an otherwise immediate event is triggered (a screen or track event) - this will result in the normally immediate event having some delay of < 50ms, as it is merged into this queue (hopefully rare).

### DIFF
--- a/Sources/AppcuesKit/Data/Analytics/AnalyticsTracker.swift
+++ b/Sources/AppcuesKit/Data/Analytics/AnalyticsTracker.swift
@@ -30,12 +30,12 @@ internal class AnalyticsTracker: AnalyticsTracking, AnalyticsSubscribing {
     private let syncQueue = DispatchQueue(label: "appcues-analytics")
 
     // this is the background analytics processing queue - 10 sec batch
-    private var pendingActivity: [Activity] = []
+    private var backgroundActivity: [Activity] = []
     // this is the immediate processing queue - 50ms batch to group identify with immediate subsequent updates
     private var priorityActivity: [Activity] = []
 
     // holds the work item for background analytics (flow events) to batch and send in 10 sec intervals
-    private var flushWorkItem: DispatchWorkItem?
+    private var backgroundFlushWorkItem: DispatchWorkItem?
 
     // holds items to be immediately processed, but allowed to group with very near additional updates (50ms)
     private var priorityFlushWorkItem: DispatchWorkItem?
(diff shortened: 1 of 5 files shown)


Sources/AppcuesKit/Data/Analytics/AnalyticsTracker.swift
```diff
@@ -28,9 +28,18 @@ internal class AnalyticsTracker: AnalyticsTracking, AnalyticsSubscribing {
 
     // maintain a batch of asynchronous events that are waiting to be flushed to network
     private let syncQueue = DispatchQueue(label: "appcues-analytics")
+
+    // this is the background analytics processing queue - 10 sec batch
     private var pendingActivity: [Activity] = []
```
iujames, line 33: I didn't rename things for simpler review in this commit, but I do think I'll make another commit pass to rename `pendingActivity` to `backgroundActivity`, and `flushWorkItem` to `backgroundFlushWorkItem` - and whatever other naming clarifications should be done. I think it would ideally be more clear why there are two code paths and exactly which case they are for.

Sources/AppcuesKit/Data/Analytics/AutoPropertyDecorator.swift
```diff
@@ -60,9 +60,7 @@ internal class AutoPropertyDecorator: AnalyticsDecorating {
             return tracking
         case .group:
             // group updates only have this single auto prop, so add that and return early
-            decorated.properties = (decorated.properties ?? [:]).merging([
-                "_lastSeenAt": Date()
-            ])
+            decorated.properties = (decorated.properties ?? [:]).merging(["_lastSeenAt": Date()])
```
iujames, line 63: make SwiftLint happy on line count in this function :)

Tests/AppcuesKitTests/Analytics/AnalyticsTrackerTests.swift
```diff
@@ -66,6 +66,123 @@ class AnalyticsTrackerTests: XCTestCase {
...
+        waitForExpectations(timeout: 1)
+    }
+
+    func testIdentifyWithScreen() throws {
+        // Arrange
+        let onRequestExpectation = expectation(description: "Valid request")
+        let expectedEvents = [Event(name: "appcues:screen_view", attributes: ["screenTitle":"screen-name"])]
+
+        // In the case of a profile update, immediately followed by a group update, verify that the
+        // updates get merged together in a single Activity that gets sent to the ActivityProcessor
```
mmaatttt, line 135: I think this comment is a copy/paste error?

mmaatttt (approved):
This looks good to me! Thank you for adding the detailed comments throughout

Tests/AppcuesKitTests/Analytics/AnalyticsTrackerTests.swift
iujames: yep! will clean up, thanks.

PR #473, opened 2023-11-20
Fix/imports

mmaatttt:
A default `import Foundation` doesn't work for Swift 5.6.

### DIFF
--- a/Sources/AppcuesKit/Presentation/Public/ElementTargeting/AppcuesViewElement.swift
+++ b/Sources/AppcuesKit/Presentation/Public/ElementTargeting/AppcuesViewElement.swift
@@ -6,7 +6,7 @@
 //  Copyright © 2023 Appcues. All rights reserved.
 //
 
-import Foundation
+import UIKit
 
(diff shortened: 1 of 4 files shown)


Sources/AppcuesKit/Presentation/Public/ElementTargeting/AppcuesViewElement.swift
```diff
@@ -6,7 +6,7 @@
 //  Copyright © 2023 Appcues. All rights reserved.
 //
 
-import Foundation
+import UIKit
```
iujames, line 9: Is this just a convenience to get CoreGraphics and Foundation? I don't think UIKit actually used here, correct?

iujames (approved):

Sources/AppcuesKit/Presentation/Public/ElementTargeting/AppcuesViewElement.swift
mmaatttt: Correct, I suppose this could be `import CoreGraphics`.

PR #474, opened 2023-11-22
Add support for step error handling and recovery

iujames:
The work here is analogous to what is being worked on the Android side in https://github.com/appcues/appcues-android-sdk/pull/528.

At a high level: an `AppcuesTraitError` can be marked as `recoverable`. A `.step` error resulting from this can then instruct the state machine to go into a new `.failing` state, rather than being considered fatal, as before. When in the `.failing` state, a `.retry` action can trigger another attempt to move to the desired target state (the state of the transition when it failed previously), and retry some `SideEffect` (as of now, always the container presentation).

If the step is able to recover and present successfully, it moves back into the normal state machine flow (.renderingStep) and an `appcues:v2:step_recovery` analytics event is generated, with a matching `errorId` to the original `step_error` that occurred.

This work will enable future anchored tooltip error/retry logic, to handle target elements that are not immediately visible on screen load. Selected `AppcuesTraitError` instances here will be recoverable - selector not found, or multiple matching selectors. Others will not be recoverable and will end the experience as before.

4 commits here
1. Initial implementation of state machine changes
2. Code cleanup from initial review, fixing lints, and getting existing tests passing
3. New tests for new functionality
4. Disabling (commenting out) the test retry logic - the actual retry logic, based on detecting UI changes that should trigger a retry, will be follow on work. This was just some WIP test code to retry after 3 seconds to prove concepts.

The end result will hopefully be support of error handling and recovery to show tooltips on items that later scroll into view, like this example:

https://github.com/appcues/appcues-ios-sdk/assets/19266448/13f8d6b0-d369-4f9f-821d-855bfa536aeb

### DIFF
--- a/Sources/AppcuesKit/Presentation/ExperienceRendering/ExperienceRenderer.swift
+++ b/Sources/AppcuesKit/Presentation/ExperienceRendering/ExperienceRenderer.swift
@@ -182,12 +182,10 @@ internal class ExperienceRenderer: ExperienceRendering, StateMachineOwning {
         // if an active experiment does exist, it should now track the experiment_entered analytic
         track(experiment: experience.experiment)
 
-        // only track analytics on published experiences (not previews)
         // and only add the observer if the state machine is idling, otherwise there's already another experience in-flight
         if stateMachine.state == .idling {
-            if experience.published {
-                stateMachine.addObserver(analyticsObserver)
-            }
+            // we always add an analytics observer, it will internally filter out unpublished flows (builder previews)
+            stateMachine.addObserver(analyticsObserver)
 
             if experience.renderContext == .modal {
                 // add recovery observer - on recoverable step errors, initiate recovery and retry
--- a/Sources/AppcuesKit/Presentation/ExperienceRendering/ExperienceRenderer.swift
+++ b/Sources/AppcuesKit/Presentation/ExperienceRendering/ExperienceRenderer.swift
@@ -61,7 +61,7 @@ internal class StepRecoveryObserver: ExperienceStateObserver {
     }
 
     func stopRetryHandler() {
-        AppcuesScrollViewDelegate.shared.observer = nil
+        AppcuesScrollViewDelegate.shared.detach()
     }
 
     func evaluateIfSatisfied(result: ExperienceStateObserver.StateResult) -> Bool {
@@ -74,7 +74,7 @@ internal class StepRecoveryObserver: ExperienceStateObserver {
     }
 
     private func startRetryHandler() {
-        AppcuesScrollViewDelegate.shared.observer = self
+        AppcuesScrollViewDelegate.shared.attach(using: self)
     }
 }
 
@@ -106,9 +106,6 @@ internal class ExperienceRenderer: ExperienceRendering, StateMachineOwning {
         self.stepRecoveryObserver = StepRecoveryObserver(stateMachine: stateMachine)
 
         stateMachines[ownerFor: .modal] = self
-
-        // TODO: guard against multiple
-        UIScrollView.swizzleScrollViewGetDelegate()
     }
 
     func start(owner: StateMachineOwning, forContext context: RenderContext) {
@@ -170,9 +167,7 @@ internal class ExperienceRenderer: ExperienceRendering, StateMachineOwning {
 
     private func show(experience: ExperienceData, completion: ((Result<Void, Error>) -> Void)?) {
         guard Thread.isMainThread else {
-            DispatchQueue.main.async {
-                self.show(experience: experience, completion: completion)
-            }
+            DispatchQueue.main.async { self.show(experience: experience, completion: completion) }
             return
         }
 
@@ -376,9 +371,7 @@ internal class ExperienceRenderer: ExperienceRendering, StateMachineOwning {
 
     func resetAll() {
         guard Thread.isMainThread else {
-            DispatchQueue.main.async {
-                self.resetAll()
-            }
+            DispatchQueue.main.async { self.resetAll() }
             return
         }
 
--- a/Sources/AppcuesKit/Presentation/ExperienceRendering/ExperienceRenderer.swift
+++ b/Sources/AppcuesKit/Presentation/ExperienceRendering/ExperienceRenderer.swift
@@ -41,43 +41,6 @@ internal enum ExperienceRendererError: Error {
     case experimentControl
 }
 
-// This class is only used by the Modal RenderContext (modals and tooltips)
-// to detect recoverable step errors and attempt to retry when scroll changes
-// are observed. It hooks into the AppcuesScrollViewDelegate, which receives
-// scroll updates from the UIScrollView implementations in the app via
-// method swizzling.
-@available(iOS 13.0, *)
-internal class StepRecoveryObserver: ExperienceStateObserver {
-
-    private let stateMachine: ExperienceStateMachine
-
-    init(stateMachine: ExperienceStateMachine) {
-        self.stateMachine = stateMachine
-    }
-
-    func scrollEnded() {
-        stopRetryHandler() // stop now, if this next retry fails, it will start over again
-        try? stateMachine.transition(.retry)
-    }
-
-    func stopRetryHandler() {
-        AppcuesScrollViewDelegate.shared.detach()
-    }
-
-    func evaluateIfSatisfied(result: ExperienceStateObserver.StateResult) -> Bool {
-        if case .failure(.step(_, _, _, recoverable: true)) = result {
-            startRetryHandler()
-        }
-
-        // recovery observer never stops observing
-        return false
-    }
-
-    private func startRetryHandler() {
-        AppcuesScrollViewDelegate.shared.attach(using: self)
-    }
-}
-
 @available(iOS 13.0, *)
 internal class ExperienceRenderer: ExperienceRendering, StateMachineOwning {
 
--- a/Sources/AppcuesKit/Presentation/ExperienceRendering/ExperienceRenderer.swift
+++ b/Sources/AppcuesKit/Presentation/ExperienceRendering/ExperienceRenderer.swift
@@ -100,7 +100,6 @@ internal class ExperienceRenderer: ExperienceRendering, StateMachineOwning {
         if shouldClearCache {
             potentiallyRenderableExperiences = [:]
             stateMachines.cleanup()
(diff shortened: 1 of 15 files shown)


Sources/AppcuesKit/Presentation/ExperienceRendering/ExperienceStateMachine+Action.swift
```diff
@@ -10,12 +10,13 @@ import Foundation
 
 @available(iOS 13.0, *)
 extension ExperienceStateMachine {
-    enum Action {
+    indirect enum Action {
```
iujames, line 13: `indirect` since new `reportError` can contain a `SideEffect`, and a `SideEffect` can contain an `Action` (i.e. `.continuation(Action)`)

Sources/AppcuesKit/Presentation/ExperienceRendering/ExperienceStateMachine+State.swift
```diff
@@ -40,9 +43,12 @@ extension ExperienceStateMachine {
                     .endingStep(_, let stepIndex, _, _),
                     .endingExperience(_, let stepIndex, _):
                 return stepIndex
+            case .failing(let targetState, _):
+                return targetState.currentStepIndex
             }
         }
 
+        // swiftlint:disable cyclomatic_complexity
```
iujames, line 51: note the disable. It seemed more maintainable in this case to keep all the transitions consolidated in the one spot still, rather than break apart.

Sources/AppcuesKit/Presentation/ExperienceRendering/ExperienceStateMachine+State.swift
```diff
@@ -102,6 +124,9 @@ extension ExperienceStateMachine.State: Equatable {
             return experience1.id == experience2.id && stepIndex1 == stepIndex2 && markComplete1 == markComplete2
         case let (.endingExperience(experience1, stepIndex1, markComplete1), .endingExperience(experience2, stepIndex2, markComplete2)):
             return experience1.id == experience2.id && stepIndex1 == stepIndex2 && markComplete1 == markComplete2
+        case let (.failing(targetState1, _), .failing(targetState2, _)):
+            // TODO: to fully compare equality, would need to make SideEffect equatable (many cascading updates)
+            return targetState1 == targetState2
```
iujames, line 129: Note the TODO - I believe this is only really used in unit tests to compare states. This one is tricky, it would require making SideEffect equatable, then making Action equatable, and perhaps several other pieces of the associated values in those enums.

Sources/AppcuesKit/Presentation/ExperienceRendering/ExperienceStateMachine+State.swift
```diff
@@ -135,11 +162,11 @@ extension ExperienceStateMachine.Transition {
     static func fromIdlingToBeginningExperience(_ experience: ExperienceData) -> Self {
         // handle errors that occurred prior to starting the experience - such as flow deserialization issues
         if let error = experience.error, !error.isEmpty {
-            return .init(toState: nil, sideEffect: .error(.experience(experience, error), reset: true))
+            return .init(toState: .idling, sideEffect: .error(.experience(experience, error)), resetObservers: true)
```
iujames, line 165: we no longer need to worry about using `nil` here and the `reset: true`, since we explicitly control observer maintenance on the transition via the new `resetObservers` flag. This allows removal of the logic to always reset on the move to `.idling` from before - as that is problematic when the new `.failing` state needs to end and process a new `.startExperience`. In general, I think it is more straightforward to handle the reset explicitly in each transition as desired, so I've updated all such spots.

Sources/AppcuesKit/Presentation/ExperienceRendering/ExperienceStateMachine+State.swift
```diff
@@ -70,15 +76,31 @@ extension ExperienceStateMachine {
...
                     sideEffect = .processActions(experience.postExperienceActionFactory)
                 }
-                return Transition(toState: .idling, sideEffect: sideEffect)
+                return Transition(toState: .idling, sideEffect: sideEffect, resetObservers: true)
+            case let (.failing(targetState, retryEffect), .retry):
+                return Transition(toState: targetState, sideEffect: retryEffect)
+            case (.failing, .startExperience):
+                // do not reset observers here, as we want to maintain the attached analytics listener
+                // through the transition to idling and on to the new experience start attempt
+                return Transition(toState: .idling, sideEffect: .continuation(action), resetObservers: false)
```
iujames, line 85: note: `resetObservers` is false here, as we are moving right on to starting a new experience and want to maintain those, I think? (the analytics observer namely)

Sources/AppcuesKit/Presentation/ExperienceRendering/ExperienceStateMachine.swift
```diff
@@ -219,9 +223,10 @@ extension ExperienceStateMachine {
     enum SideEffect {
         case continuation(Action)
         case presentContainer(ExperienceData, Experience.StepIndex, ExperiencePackage, [Experience.Action])
+        case retryPresentation(ExperienceData, Experience.StepIndex, ExperiencePackage)
```
iujames, line 226: this _could_ use the existing `.presentContainer`, but I preferred the new `.retryPresentation` to be explicit. It (1) avoids adding another `isRecovering` in the associated values of the existing enum value (which is a swiftlint violation) and (2) makes it clear on L252 below that we are in recovery mode, which is needed on L288 to not add duplicate pageMonitor observers.

Sources/AppcuesKit/Presentation/ExperienceRendering/ExperienceStateMachine.swift
```diff
@@ -312,10 +332,26 @@ extension ExperienceStateMachine {
...
+
+        private func handlePresentationError(
+            _ error: Error,
+            machine: ExperienceStateMachine,
+            experience: ExperienceData,
+            stepIndex: Experience.StepIndex,
+            package: ExperiencePackage
+        ) {
+            let recoverable = (error as? AppcuesTraitError)?.recoverable ?? false
+            let retryEffect: ExperienceStateMachine.SideEffect? = recoverable ? .retryPresentation(experience, stepIndex, package) : nil
```
iujames, line 348: sort of subtle here, but a `nil` retryEffect basically means it was a fatal error and things conclude. If non-nil, it will go into `.failure` mode and hope for recovery.

Sources/AppcuesKit/Presentation/ExperienceRendering/ExperienceRenderer.swift
```diff
@@ -97,6 +148,7 @@ internal class ExperienceRenderer: ExperienceRendering, StateMachineOwning {
         if shouldClearCache {
             potentiallyRenderableExperiences = [:]
             stateMachines.cleanup()
+            stepRecoveryObserver.stopRetryHandler()
```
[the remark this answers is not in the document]
iujames: hey @mmaatttt , checking understanding - do you mean move this down into `dismiss(inContext..)` down around L329 in this file? so a check there that the context is the modal one, then stopping any retry? I think that would work, only other way a dismiss would occur is if the VC is dismissed in the view layer and that would not occur anyway for an experience sitting in the error state.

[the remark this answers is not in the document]
mmaatttt: Yeah, that's what I was thinking. Just an idea, I haven't thought about this half as much as you have. I think it'd be exceedingly rare that the way it is now could cause a problem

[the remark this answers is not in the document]
iujames: ok what I think might be most appropriate is to handle this in the observer itself, with something like this `.idling` check (will clean up the code into a single switch)

```swift
    func evaluateIfSatisfied(result: ExperienceStateObserver.StateResult) -> Bool {
        if case .failure(.step(_, _, _, recoverable: true)) = result {
            startRetryHandler()
        }

        // if the machine goes back to idling, this means that any experience
        // that was in a retry state was fully dismissed, or potentially a new
        // experience has been queued up to start in its place - remove any
        // existing retry handler to stop attempting recovery
        if case .success(.idling) = result {
            stopRetryHandler()
        }

        // recovery observer never stops observing
        return false
    }
```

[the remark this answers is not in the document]
mmaatttt: Yeah, nice work!

Sources/AppcuesKit/Presentation/ExperienceRendering/ExperienceStateMachine+State.swift
iujames: thats a good edge case catch, yes, it would. Need to think about how to handle that one. We don't have that issue on Android, as the check for published flow happens in a different spot inside the observer itself. But here, we'll need something to handle it.

Sources/AppcuesKit/Presentation/ExperienceRendering/ExperienceStateMachine+State.swift
```diff
@@ -70,15 +76,31 @@ extension ExperienceStateMachine {
...
+            case (.failing, .endExperience):
+                return Transition(toState: .idling, resetObservers: true)
```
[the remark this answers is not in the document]
iujames: Yes, that was the intention there - an error ending an experience sends no further analytics. I believe this matches how it was before, only it was an `experience_error` before and now it is a `step_error`.

mmaatttt (approved):
Updated state machine diagram. New stuff in red.

Mobile Experience States@2x

Sources/AppcuesKit/Presentation/ExperienceRendering/ExperienceStateMachine+State.swift
iujames: left this code as-is and updated in https://github.com/appcues/appcues-ios-sdk/pull/474/commits/57534732108e1c72635d662faae022491d6835d5 to always attach an analytics observer, but have it filter internally to avoid reporting analytics for any unpublished flow content (matching Android behavior)

mmaatttt: Perfect!

PR #475, opened 2023-11-22
Ensure resetting embed frames happens on main thread

mmaatttt:
Noticed this bug in the showcase app. Signing out (ie calling `Appcues.reset()`) would crash, because it was trying to do UI operations on a background thread. The operations in question are resetting embed frames. So I've added our standard main thread check to `ExperienceRenderer.resetAll()` and that fixes it.


iujames (approved):
nice catch!

PR #483, opened 2023-12-18
Guard against an invalid view rect in screen capture

iujames:
Discovered in testing / prepping a demo - in some cases an invalid intersection rect can be returned in the check with the safe area bounds, resulting in a layout object with values that cannot be serialized to JSON correctly, and fail upload. This new check will prevent that.

### DIFF
--- a/Sources/AppcuesKit/Presentation/Debugger/ScreenCapturing/UIKitElementTargeting.swift
+++ b/Sources/AppcuesKit/Presentation/Debugger/ScreenCapturing/UIKitElementTargeting.swift
@@ -192,18 +192,25 @@ internal extension UIView {
             return subview.asViewElement(in: bounds, safeAreaInsets: childInsets, autoTag: childAutoTag)
         }
 
-        // only create a selector for elements that have at least the center point 
-        // visible in the current screen bounds, inset by any safe area adjustments
+        // find the rect of the visible area of the view within the safe area
         let safeBounds = bounds.inset(by: safeAreaInsets)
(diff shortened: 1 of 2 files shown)


Sources/AppcuesKit/Presentation/Debugger/ScreenCapturing/UIKitElementTargeting.swift
```diff
@@ -192,11 +192,15 @@ internal extension UIView {
...
         }
 
-        // only create a selector for elements that have at least the center point 
-        // visible in the current screen bounds, inset by any safe area adjustments
+        // find the rect of the visible area of the view within the safe area
         let safeBounds = bounds.inset(by: safeAreaInsets)
-        let centerPointVisible = safeBounds.contains(CGPoint(x: absolutePosition.midX, y: absolutePosition.midY))
         let visibleRect = safeBounds.intersection(absolutePosition)
+        // if there is no intersection of this view in the safe area, it is skipped
+        guard !visibleRect.isNull else { return nil }
```
iujames, line 199: only question I have is if this opens some edge case where a parent is not in the safe area but it's children are (outside its bounds)?

iujames: reworked this - now it will not skip the element, so any children are still potentially captured in selectors. There will not be a selector generated for the non-visible element, since it's center point is not in the safe bounds (L207 check)

mmaatttt (approved):

PR #488, opened 2024-01-11
✨ Adding config.attributes to action @appcues/track

andretortolano:
This is part of a larger initiative for integrations related to button clicks, on our end for now we need to add an attribute property that can receive generic information from builder, and pass it as properties when calling appcues.track internally.

### DIFF
--- a/Sources/AppcuesKit/Presentation/Actions/Appcues/AppcuesTrackAction.swift
+++ b/Sources/AppcuesKit/Presentation/Actions/Appcues/AppcuesTrackAction.swift
@@ -10,8 +10,9 @@ import Foundation
 
 @available(iOS 13.0, *)
 internal class AppcuesTrackAction: AppcuesExperienceAction {
-    struct Config: Decodable {
+    struct Config {
         let eventName: String
+        let attributes: [String: Any]?
     }
 
     static let type = "@appcues/track"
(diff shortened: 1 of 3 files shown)


Tests/AppcuesKitTests/Actions/AppcuesTrackActionTests.swift
```diff
@@ -48,6 +48,29 @@ class AppcuesTrackActionTests: XCTestCase {
...
+    
+    func testExecuteWithAttributes() throws {
+        // Arrange
+        var completionCount = 0
+        var trackCount = 0
+        appcues.analyticsPublisher.onPublish = { trackingUpdate in
+            XCTAssertEqual(trackingUpdate.type, .event(name: "My Custom Event", interactive: true))
+            
+            [
+                "_builderButtonClick": true,
```
andretortolano, line 60: can replace this with:

```
"boolean": true,
                                            "string": "string",
                                            "int": 10,
                                            "double": 10.5
```

to cover more lines of code if you want, did not notice it in the first pass

Sources/AppcuesKit/Presentation/Actions/Appcues/AppcuesTrackAction.swift
```diff
@@ -10,27 +10,67 @@ import Foundation
...
+        let container = try decoder.container(keyedBy: DynamicCodingKeys.self)
+
+        self.eventName = try container.decode(String.self, forKey: DynamicCodingKeys(stringValue: "eventName")!)
+        
+        var dict: [String: Any] = [:]
+
+        container.allKeys.forEach { key in
+            if key.stringValue == "eventName" {
+                // If it's the eventName, skip it
+                return
```
iujames, line 58: "eventName" is already decoded on L51, I think there is some way to just pull out "attributes" as its own container, then loop through its items?

andretortolano: actually this is probably not right. I will take another look!

Sources/AppcuesKit/Presentation/Actions/Appcues/AppcuesTrackAction.swift
```diff
@@ -10,27 +10,67 @@ import Foundation
...
+extension AppcuesTrackAction.Config: Decodable {
+    // Custom decoding for this one - we want to just gather up all the key/value
+    // pairs from the trait config as raw user property values. We trim the set to
+    // only those of supported data types and then store in the resulting properties
+    // dictionary
+    init(from decoder: Decoder) throws {
+               
+        let container = try decoder.container(keyedBy: DynamicCodingKeys.self)
+
+        self.eventName = try container.decode(String.self, forKey: DynamicCodingKeys(stringValue: "eventName")!)
```
mmaatttt, line 51: FYI instead of `DynamicCodingKeys(stringValue: "eventName")!`, you can use `DynamicCodingKeys(key: String)` to avoid the force unwrap. If you look at the implementation, it does the same thing, but just isn't an optional init since the optional init is only required by the `CodingKey` contract.

And technically since `eventName` and `attributes` are known keys, you should probably do this instead:
```swift
enum CodingKeys: CodingKey {
    case eventName
    case attributes
}
...
let container = try decoder.container(keyedBy: CodingKeys.self)
self.eventName = try container.decode(String.self, forKey: .eventName)
```

Sources/AppcuesKit/Presentation/Actions/Appcues/AppcuesTrackAction.swift
mmaatttt: If you use the CodingKey approach I suggested above:

```swift
var attributesContainer = container.nestedContainer(keyedBy: DynamicCodingKeys.self, forKey: .attributes)
```

Tests/AppcuesKitTests/Actions/AppcuesTrackActionTests.swift
mmaatttt: Good idea!

Sources/AppcuesKit/Presentation/Actions/Appcues/AppcuesTrackAction.swift
andretortolano: working through that!

Sources/AppcuesKit/Presentation/Actions/Appcues/AppcuesTrackAction.swift
```diff
@@ -10,27 +10,68 @@ import Foundation
...
-    struct Config: Decodable {
+    struct Config {
         let eventName: String
+        let attributes: [String: Any]?
+    }
+    
+    private enum CodingKeys: String, CodingKey {
+           case eventName
+           case attributes
     }
```
mmaatttt, line 21: Minor organization nitpick: I'd move this down inside the extension on L48.

Then you can reference `CodingKeys.self` instead of `AppcuesTrackAction.CodingKeys.self`

mmaatttt (approved):
Nice work!

PR #489, opened 2024-01-17
Support additional query parameters on deep links to preview or show experiences

iujames:
Proposing an idea here, not sold on it - but this one would allow us to take any set of additional query parameters on the `/experience_preview` or `/experience_content` SDK deep links. This would be used to support a new `?locale_id={id}` parameter during the QR code preview flow from the builder. This approach is more generic, where it would just take any set of params given and pass them through, verbatim, to the api request that gets made.

The more specific approach would be to simply read the one `locale_id` param on the preview call directly, and use that in only the preview call. I thought perhaps a more general approach here would let us possibly add other things in the future without SDK changes. The more I think about it, it seems like YAGNI and I'm not happy with the change amplification here really. Interested in thoughts. Easy to update to a more targeted approach.

### DIFF
--- a/Tests/AppcuesKitTests/Experiences/ExperienceLoaderTests.swift
+++ b/Tests/AppcuesKitTests/Experiences/ExperienceLoaderTests.swift
@@ -39,7 +39,7 @@ class ExperienceLoaderTests: XCTestCase {
         let completionExpectation = expectation(description: "Completion called")
 
         // Act
-        experienceLoader.load(experienceID: "123", published: true, trigger: .showCall) { result in
+        experienceLoader.load(experienceID: "123", published: true, queryItems: [], trigger: .showCall) { result in
             if case .success = result {
                 completionExpectation.fulfill()
             }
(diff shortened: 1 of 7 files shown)


Tests/AppcuesKitTests/Experiences/ExperienceLoaderTests.swift
```diff
@@ -68,7 +68,7 @@ class ExperienceLoaderTests: XCTestCase {
         let completionExpectation = expectation(description: "Completion called")
 
         // Act
-        experienceLoader.load(experienceID: "123", published: false, trigger: .preview) { result in
+        experienceLoader.load(experienceID: "123", published: false, queryItems: [], trigger: .preview) { result in
```
mmaatttt, line 71: Maybe add a `queryItem` here and then explicitly check it was added properly up on L55?

mmaatttt (approved):
I like this generic approach. I don't think it's meaningfully more complex than working with just a parameter for locale.

Do we need query items for the `Action.show` case? IIRC we determined that the backend needs an update to support this but I don't think we need to pass the locale here? So is this just general future proofing?

PR #490, opened 2024-01-22
Preserve debug data through debugger mode changes

mmaatttt:
This allows the debugger to inspect the screen capture flow network requests, which will make troubleshooting that flow much easier. Added bonus is that `UIDebugger.show(mode:)` is now simpler overall!

Instead of calling `hide()` and tearing down and recreating the debug panel and data when the mode changes, we preserve it, which means we can capture the data flow through the screen capture process. A few changes were needed in `DebugViewController` to allow `mode` to be changed and update the view to reflect the new mode instead of just being a `let` constant that is configured at init.


iujames (approved):
nice upgrade!

PR #498, opened 2024-02-16
Add _device auto properties

mmaatttt:
Did a bit of cleanup and then added the initial `_device` structure. Also updated the debugger to show `_device` as a standalone section.

### DIFF
--- a/Sources/AppcuesKit/Data/Analytics/AutoPropertyDecorator.swift
+++ b/Sources/AppcuesKit/Data/Analytics/AutoPropertyDecorator.swift
@@ -22,7 +22,23 @@ internal class AutoPropertyDecorator: AnalyticsDecorating {
     private let config: Appcues.Config
 
     // these are the fixed values for the duration of the app runtime
-    private var applicationProperties: [String: Any] = [:]
+    private lazy var applicationProperties: [String: Any] = {
+        [
+            "_appId": config.applicationID,
+            "_operatingSystem": "iOS",
+            "_bundlePackageId": Bundle.main.identifier,
+            "_appName": Bundle.main.displayName,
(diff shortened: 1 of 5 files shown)


Sources/AppcuesKit/Data/Analytics/AutoPropertyDecorator.swift
```diff
@@ -39,10 +39,23 @@ internal class AutoPropertyDecorator: AnalyticsDecorating {
     }
 
     func decorate(_ tracking: TrackingUpdate) -> TrackingUpdate {
+        // Early return for group updates
```
mmaatttt, line 42: I found it confusing that there was returns mixed in below, so moving those to the top.

Sources/AppcuesKit/Data/Analytics/AutoPropertyDecorator.swift
```diff
@@ -55,38 +68,28 @@ internal class AutoPropertyDecorator: AnalyticsDecorating {
...
-        var sessionProperties: [String: Any?] = [
+        let sessionProperties: [String: Any?] = [
             "userId": storage.userID,
             "_isAnonymous": storage.isAnonymous,
             "_localId": storage.deviceID,
             "_sessionPageviews": sessionPageviews,
             "_sessionRandomizer": sessionRandomizer,
             "_currentScreenTitle": currentScreen,
             "_lastScreenTitle": previousScreen,
+            "_lastBrowserLanguage": Bundle.main.preferredLocalizations.first,
```
mmaatttt, line 85: If this value is `nil` it gets filtered out by `compactMapValues` below, so no change in behaviour, and now `sessionProperties` can be `let`.

iujames (approved):
🚀

PR #500, opened 2024-02-20
Improve network logging

mmaatttt:
Ensure that a helpful response is always logged. Previously if there was no HTTP status code, we'd not get a log which makes debugging extra tricky. Now we always log _something_, and add the `error` body and empty message if those conditions would cause the response decode to fail.

Also, in the case where `data == nil`, we were returning without calling the completion handler, which, while unlikely, is bad, so I added an error completion block there too.

Now we can get something like this, instead of nothing, which is a big improvement.

Simulator Screenshot - iPhone 13 Pro - 2024-02-20 at 16 39 16


iujames (approved):
handy!

andretortolano:
500!

PR #501, opened 2024-02-21
Centralize analytic event names

mmaatttt:
This is some refactoring targeting `main` before adding the new device events in the `sdk4` branch. Basically, I don't like that currently,
1) event names are scattered in different places (manifests the question, "where do I put the new device event names"), and
2) the name `LifecycleEvent` always confused me when I was looking for them.

Now there's `enum Events` which namespaces `Session` and `Experience` (and soon `Device`). This seems substantially more friendly to anyone new to the codebase.

Also in the name of clarity, I replaced the strange static extension `LifecycleEvent.properties()` with a `Dictionary` initializer.

Last little change was renaming `func trackLifecycleEvent(_ name: LifecycleEvent, _ properties: [String: Any])` to `func track(experienceEvent name: Events.Experience, properties: [String: Any])` in `ExperienceStateObserver`.

### DIFF
--- a/Sources/AppcuesKit/Presentation/Debugger/Panel/LoggedEvent.swift
+++ b/Sources/AppcuesKit/Presentation/Debugger/Panel/LoggedEvent.swift
@@ -87,10 +87,10 @@ internal struct LoggedEvent: Identifiable {
         self.properties = update.properties
 
         switch update.type {
-        case let .event(name, _) where SessionEvents.allNames.contains(name):
+        case let .event(name, _) where Events.Session(rawValue: name) != nil:
             self.type = .session
             self.name = name.prettifiedEventName
-        case let .event(name, _) where name.starts(with: "appcues:v2:"):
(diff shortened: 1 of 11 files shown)


Sources/AppcuesKit/Presentation/Debugger/Panel/LoggedEvent.swift
```diff
@@ -87,10 +87,10 @@ internal struct LoggedEvent: Identifiable {
         self.properties = update.properties
 
         switch update.type {
-        case let .event(name, _) where SessionEvents.allNames.contains(name):
+        case let .event(name, _) where Events.Session(rawValue: name) != nil:
             self.type = .session
             self.name = name.prettifiedEventName
-        case let .event(name, _) where name.starts(with: "appcues:v2:"):
+        case let .event(name, _) where Events.Experience(rawValue: name) != nil:
```
iujames, line 93: nice improvement here to not rely on this `v2` nuance

iujames (approved):
nice clean up!

PR #503, opened 2024-02-23
Ensure actions with analytics log instead of publish for experience previews

mmaatttt:
1. Update `AnalyticsPublisher` to be able to conditionally publish or debug log analytics
    - `decorateAndPublish()` got split into `decorate()` and `notifySubscribers()` so that the logging path can also decorate
2. Log experience analytics for flow previews instead of doing nothing
3. Log action analytics with `@appcues/step_interaction` and `@appcues/submit-form` for flow previews instead of publishing
4. Add a share sheet link to the debugger to allow easy sharing (or copying) the entire debug log
5. New tests

Simulator Screenshot - iPhone 15 Pro - 2024-02-23 at 09 52 59

### DIFF
--- a/Sources/AppcuesKit/Presentation/ExperienceRendering/ExperienceStateObserver.swift
+++ b/Sources/AppcuesKit/Presentation/ExperienceRendering/ExperienceStateObserver.swift
@@ -69,11 +69,6 @@ extension ExperienceStateMachine {
         }
 
         func evaluateIfSatisfied(result: ExperienceStateObserver.StateResult) -> Bool {
-            guard result.shouldTrack else {
-                // Always continue observing
-                return false
-            }
-
             switch result {
             case .success(.idling):
                 break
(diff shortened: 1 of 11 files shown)


Sources/AppcuesKit/Presentation/ExperienceRendering/ExperienceStateObserver.swift
```diff
@@ -184,19 +179,19 @@ extension ExperienceStateMachine {
...
-            return error.shouldTrack
+            return error.shouldPublish
         }
     }
 }
 
 @available(iOS 13.0, *)
 private extension ExperienceStateMachine.ExperienceError {
-    var shouldTrack: Bool {
+    var shouldPublish: Bool {
```
mmaatttt, line 194: Worth noting that now a `.noTransition` error will show up in the debugger too which would hopefully make that state machine debugging scenario easier to pick out.

iujames (approved):
nice improved handling of these!

PR #504, opened 2024-02-26
Add push status device properties

mmaatttt:
I created a new `PushMonitor` dependency, mostly because of the requirement to watch for new foreground events—it felt weird to have that happening in the `AutoPropertyDecorator`, or anywhere else non-specific. Push handlers and other stuff might end up in `PushMonitor` (and necessitate a rename) in the future, but I haven't really planned that out yet.

The non-straightforward part of this is that calls to `UNUserNotificationCenter` crash in tests for non-app targets ([Stack Overflow](https://stackoverflow.com/questions/77448766/app-crashes-when-im-getting-unusernotificationcenter-current-from-into-a-swif)). Because there are tests that use `Appcues` in addition to `MockAppcues`, we can't use `UNUserNotificationCenter` without doing something different in the SDK code when tests are running.

I started off by creating a protocol to be able to mock the notification center, but I didn't like that the mock class had to be in the core SDK code (even if I could `#if DEBUG` wrap it). I've settled (for now) on an approach that has the least test-specific code in the SDK (1 line guard, and a 1 line testing function). Note that I could remove `mockPushStatus()` and just make `var pushAuthorizationStatus` non-private to be able to set it directly in tests, but that could introduce confusion in the future. Other options would be optional dependencies that we know are resolved in non-debug builds, but I prefer the compile-time guarantee of my current approach.

Also [sc-60921]

### DIFF
--- a/Sources/AppcuesKit/Appcues.swift
+++ b/Sources/AppcuesKit/Appcues.swift
@@ -346,6 +346,7 @@ public class Appcues: NSObject {
         container.registerLazy(ActivityProcessing.self, initializer: ActivityProcessor.init)
         container.registerLazy(ActivityStoring.self, initializer: ActivityFileStorage.init)
         container.registerLazy(AnalyticsBroadcaster.self, initializer: AnalyticsBroadcaster.init)
+        container.registerLazy(PushMonitoring.self, initializer: PushMonitor.init)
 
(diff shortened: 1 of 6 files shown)


iujames (approved):
nice clean split of responsibilities with the PushMonitor 👍

PR #505, opened 2024-02-27
Add push primer support

mmaatttt:
New `_pushPrimerEligible` auto prop and `@appcues/request-push` action.

Also [sc-60936]

### DIFF
--- a/Sources/AppcuesKit/Data/PushMonitor.swift
+++ b/Sources/AppcuesKit/Data/PushMonitor.swift
@@ -11,6 +11,9 @@ import UIKit
 internal protocol PushMonitoring: AnyObject {
     var pushEnabled: Bool { get }
     var pushBackgroundEnabled: Bool { get }
+    var pushPrimerEligible: Bool { get }
+
+    func refreshPushStatus(completion: ((UNAuthorizationStatus) -> Void)?)
 }
 
 internal class PushMonitor: PushMonitoring {
@@ -25,6 +28,10 @@ internal class PushMonitor: PushMonitoring {
 
     var pushBackgroundEnabled: Bool {
         storage.pushToken != nil
+    }
+
+    var pushPrimerEligible: Bool {
+        pushAuthorizationStatus == .notDetermined
     }
 
     init(container: DIContainer) {
@@ -45,14 +52,18 @@ internal class PushMonitor: PushMonitoring {
         refreshPushStatus()
     }
 
-    private func refreshPushStatus() {
+    func refreshPushStatus(completion: ((UNAuthorizationStatus) -> Void)? = nil) {
         // Skip call to UNUserNotificationCenter.current() in tests to avoid crashing in package tests
         #if DEBUG
-        guard ProcessInfo.processInfo.environment["XCTestBundlePath"] == nil else { return }
+        guard ProcessInfo.processInfo.environment["XCTestBundlePath"] == nil else {
+            completion?(pushAuthorizationStatus)
+            return
+        }
         #endif
 
         UNUserNotificationCenter.current().getNotificationSettings { [weak self] settings in
             self?.pushAuthorizationStatus = settings.authorizationStatus
+            completion?(settings.authorizationStatus)
         }
     }
 
--- a/Sources/AppcuesKit/Presentation/Actions/Appcues/AppcuesRequestPushAction.swift
+++ b/Sources/AppcuesKit/Presentation/Actions/Appcues/AppcuesRequestPushAction.swift
@@ -1,0 +1,58 @@
+//
+//  AppcuesRequestPushAction.swift
+//  AppcuesKitTests
+//
+//  Created by Matt on 2024-02-26.
+//  Copyright © 2024 Appcues. All rights reserved.
+//
+
+import Foundation
+import UserNotifications
+
+@available(iOS 13.0, *)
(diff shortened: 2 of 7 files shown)


Sources/AppcuesKit/Presentation/Actions/Appcues/AppcuesRequestPushAction.swift
```diff
@@ -0,0 +1,58 @@
...
+
+        // Skip call to UNUserNotificationCenter.current() in tests to avoid crashing in package tests
+        #if DEBUG
+        guard ProcessInfo.processInfo.environment["XCTestBundlePath"] == nil else {
+            resultHandler(false, AppcuesTraitError(description: "Executing in test environment"))
+            return
+        }
+        #endif
+
+        let options: UNAuthorizationOptions = [.alert, .sound, .badge]
```
mmaatttt, line 55: Any reason we'd want these options configurable?

Sources/AppcuesKit/Presentation/Actions/Appcues/AppcuesRequestPushAction.swift
```diff
@@ -0,0 +1,58 @@
...
+            if let error = error {
+                appcues.config.logger.error("@appcues/request-push failed with %{public}@", error.localizedDescription)
+            }
+
+            let pushMonitor = appcues.container.resolve(PushMonitoring.self)
+            let analyticsPublisher = appcues.container.resolve(AnalyticsPublishing.self)
+
+            pushMonitor.refreshPushStatus { _ in
+                analyticsPublisher.publish(TrackingUpdate(
+                    type: .event(name: Events.Device.deviceUpdated.rawValue, interactive: false),
```
mmaatttt, line 39: I think we want an `appcues:device_updated` event here (so that the `_device` profile gets an updated `_pushEnabled` value), but that wasn't made explicit anywhere.

Sources/AppcuesKit/Presentation/Actions/Appcues/AppcuesRequestPushAction.swift
iujames: I'm thinking no? If they really want to do something custom they can code the request in their own app and use the analytics listener to trigger it on an event. There is the newer `.provisional` thing. That one is silent and I guess you would also just code that into your app instead of using an action to prompt like this.

Sources/AppcuesKit/Data/PushMonitor.swift
```diff
@@ -45,14 +52,18 @@ internal class PushMonitor: PushMonitoring {
         refreshPushStatus()
     }
 
-    private func refreshPushStatus() {
+    func refreshPushStatus(completion: ((UNAuthorizationStatus) -> Void)? = nil) {
```
iujames, line 55: wondering if we need to allow some public interface to this as well, or at least call when the device token is passed in? Thinking that the `pushPrimerEligible` value could change outside of our scope, from the host app doing some permission request on its own, and we'd want to update our value so they don't later trigger a flow that doesn't make sense any longer.

Sources/AppcuesKit/Presentation/Actions/Appcues/AppcuesRequestPushAction.swift
```diff
@@ -0,0 +1,58 @@
...
+        // Skip call to UNUserNotificationCenter.current() in tests to avoid crashing in package tests
+        #if DEBUG
+        guard ProcessInfo.processInfo.environment["XCTestBundlePath"] == nil else {
+            resultHandler(false, AppcuesTraitError(description: "Executing in test environment"))
+            return
+        }
+        #endif
+
+        let options: UNAuthorizationOptions = [.alert, .sound, .badge]
+        UNUserNotificationCenter.current().requestAuthorization(options: options, completionHandler: resultHandler)
```
iujames, line 56: I forget, too much context switching, does this automatically trigger the token to get requested and sent back into the app delegate in didRegisterForRemoteNotificationsWithDeviceToken when permission is accepted?

Sources/AppcuesKit/Data/PushMonitor.swift
mmaatttt: Good question, it's a possibility, but not super likely? We can definitely expose something in the public API, but I don't really want a giant list of, "make sue you call these 20 functions in these 20 spots" to make everything integrate seamlessly.

Sources/AppcuesKit/Presentation/Actions/Appcues/AppcuesRequestPushAction.swift
mmaatttt: No, that's entirely separate and dependent on `UIApplication.registerForRemoteNotifications()`. I was wondering if we wanted to call that in this code, but I figured our install instructions would have that call in the startup sequence so it'd be redundant here.

mmaatttt: No, because `UIApplication.registerForRemoteNotifications()` can happen before the notification authorization. This is what `_pushBackgroundEnabled` would track. Then, once authorized here, `_pushEnabled = true`.

iujames: ah right, thank you.

PR #509, opened 2024-03-04
Add AppcuesPresentationDelegate with experience metadata

mmaatttt:
## Background

We provide `AppcuesExperienceDelegate` to allow apps visibility into when the presentation lifecycle of Appcues experiences. This protocol was designed before embeds (and therefore multiple render contexts) and so is no longer particularly useful for an app with multiple embed frames in addition to modals/tooltips.

## Change

I've created `AppcuesPresentationDelegate` that contains the same functions, but with a `metadata` value that provides a dictionary with `id`, `name`, and `renderContext` (`"modal" or $frameName`) to provide more clarity into what experience is triggering the presentation lifecycle methods.

## Frame delegate

I've also added `presentationDelegate` as a property on `AppcuesFrameView`. It functions exactly the same way, but is scoped to only experiences rendering in the specific frame. This doesn't really provide any new functionality since the `metadata["renderContext"]` from the global `Appcues.presentationDelegate` would allow you to filter, but it creates a better developer experience because each frame view can have its own delegate logic that can be implemented close to the UI code for the frame itself instead of once globally.

## Notes

1. For the sake of backwards compatibility, `AppcuesExperienceDelegate` remains untouched, so `AppcuesPresentationDelegate` is solely additive. We can remove `AppcuesExperienceDelegate` in SDK v4 if we so choose. I've updated the docs to indicate using `presentationDelegate` is to be preferred over `experienceDelegate`.
2. Having `AppcuesPresentationDelegate` be objc compatible creates some limitations, so `AppcuesPresentationMetadata` has to be a class instead of the struct which would be preferred.
3. `AppcuesPresentationMetadata` can be enhanced in the future with additional values if we see a need. Note that removing/renaming a value would be a breaking change.

### DIFF
--- a/Sources/AppcuesKit/Presentation/Public/AppcuesPresentationDelegate.swift
+++ b/Sources/AppcuesKit/Presentation/Public/AppcuesPresentationDelegate.swift
@@ -1,0 +1,54 @@
+//
+//  AppcuesPresentationDelegate.swift
+//  AppcuesKit
+//
+//  Created by Matt on 2024-03-04.
+//  Copyright © 2024 Appcues. All rights reserved.
+//
+
+import Foundation
+
+/// Metadata information for experiences in ``AppcuesPresentationDelegate``.
+@objc
+public class AppcuesPresentationMetadata: NSObject {
+    /// Appcues ID of the experience.
+    public let id: String
+
+    /// Name of the experience.
+    public let name: String
+
+    /// Context of the experience.
+    ///
+    /// Either `"modal"` or the `frameID` of an ``AppcuesFrame``.
+    public let renderContext: String
+
+    internal init(id: String, name: String, renderContext: String) {
+        self.id = id
+        self.name = name
+        self.renderContext = renderContext
+    }
+}
+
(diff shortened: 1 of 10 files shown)


Sources/AppcuesKit/Presentation/Public/AppcuesPresentationDelegate.swift
```diff
@@ -0,0 +1,62 @@
...
+    /// Name of the experience.
+    public let name: String
+
+    /// True if the experience is presented as an overlay. Includes modals, tooltips, and slideouts.
+    public let isOverlay: Bool
+
+    /// Context of the experience.
+    ///
+    /// Either `"modal"` or the `frameID` of an ``AppcuesFrame``.
+    public let renderContext: String
```
iujames, line 26: no real harm to have both, little redundant? other idea would be to have `isEmbedded` and then an optional `frameID`... just thinking out loud about what the consumer might comprehend most easily

iujames (approved):
looking good! nice enhancement for more detailed control if needed

PR #510, opened 2024-03-05
Add push response handler

mmaatttt:
## Notes

1. I moved things to a `Push` directory, so to see the diff, look at the 2nd commit (or 3rd for the tests).
1. The `ParsedNotification` struct takes all the [expected keys](https://www.notion.so/appcues/Push-Notification-Payload-dd5e2423c2164f7196b208bf5934f567?pvs=4#8e0251708c1b4d71b349d5a69c19cd66) from the push payload and fails otherwise. This validates the push is an Appcues one and gives us typed data to work with.
1. I added a push experience trigger instead of reusing an existing one.
1. I'm using the experience Actions primarily for code reuse (in particular the link actions handles delegates and universal links, and copy/pasting that doesn't make sense).
1. I'm not sure about the handling around sessions and user ID matching, but I've done what [we discussed](https://appcues.slack.com/archives/C031RLGS4F8/p1709652676343119) and adde tests for it.
1. There's also an [async](https://developer.apple.com/documentation/usernotifications/unusernotificationcenterdelegate/usernotificationcenter(_:didreceive:withcompletionhandler:)) version of the `UNUserNotificationCenterDelegate` method: `func userNotificationCenter(_ center: UNUserNotificationCenter, didReceive response: UNNotificationResponse) async`, so we might want to add our own async handler (using [`withCheckedContinuation`](https://developer.apple.com/documentation/swift/withcheckedcontinuation(function:_:))) for a better developer experience. 

## Usage

```swift
extension AppDelegate: UNUserNotificationCenterDelegate {
    // Asks the delegate to process the user's response to a delivered notification.
    func userNotificationCenter(
        _ center: UNUserNotificationCenter,
        didReceive response: UNNotificationResponse,
        withCompletionHandler completionHandler: @escaping () -> Void
    ) {

        if Appcues.shared.didReceiveNotification(response: response, completionHandler: completionHandler) {
            return
        }

        // Manually call completionHandler, because Appcues won't if it doesn't handle the notification
        completionHandler()
    }
}
```

Also [sc-61014]


Sources/AppcuesKit/Push/PushMonitor.swift
```diff
@@ -0,0 +1,132 @@
...
+        guard parsedNotification.userID == storage.userID else {
+            completionHandler()
+            return true
+        }
+
+        // If no session, start one for the user in the notification
+        if !appcues.isActive {
+            storage.userID = parsedNotification.userID
+            storage.isAnonymous = false
+        }
```
iujames, line 96: im internally questioning this one, seems like it could lead to some weird out of sync issue with the appcues user and the logged in user in the app potentially, in a situation where the app intentionally did not identify that user for some reason - maybe safest to just handle this right alongside L87 - user didnt match so bail out

Sources/AppcuesKit/Push/PushMonitor.swift
```diff
@@ -0,0 +1,132 @@
...
+
+        let analyticsPublisher = appcues.container.resolve(AnalyticsPublishing.self)
+        analyticsPublisher.publish(TrackingUpdate(
+            type: .event(name: Events.Push.pushOpened.rawValue, interactive: false),
+            properties: [
+                "notification_id": parsedNotification.notificationID
```
iujames, line 102: lets add `workflow_id`, `workflow_task_id` and `transaction_id` at least to start here

there are some push notification events defined here https://www.notion.so/appcues/Workflow-Analytics-Event-Schema-48b243655c394446b665ac8f759476fd?pvs=4 (scroll down a bit)

the other items should already be in there by default - account, user, device

iujames (approved):
looking great! some minor thoughts

Sources/AppcuesKit/Push/ParsedNotification.swift
```diff
@@ -0,0 +1,49 @@
...
+    let attachmentType: String?
+    let isTest: Bool
+
+    init?(userInfo: [AnyHashable: Any]) {
+        guard let accountID = userInfo["appcues_account_id"] as? String,
+        let userID = userInfo["appcues_user_id"] as? String,
+        let notificationID = userInfo["appcues_notification_id"] as? String,
+        let workflowID = userInfo["appcues_workflow_id"] as? String,
+        let workflowTaskID = userInfo["appcues_workflow_task_id"] as? String,
+        let transactionID = userInfo["appcues_transaction_id"] as? String else {
```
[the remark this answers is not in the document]
mmaatttt: Cool, makes sense. The notion doc didn't have them marked optional and I was going off that.

Sources/AppcuesKit/Push/PushMonitor.swift
mmaatttt: I agree with taking the simpler approach for now. This was provoking all kinds of red flag questions, like "what if the user actually is anonymous?" and "what about identity verification?"

Sources/AppcuesKit/Push/PushMonitor.swift
mmaatttt: Do we want to send the device auto props with `appcues:push_opened`? That's the only way `_deviceId` would make it into the payload.

iujames: the `_localId` in auto props is technically the same as `_deviceId`, but maybe yeah just add `device_id` right here as well to make it very clear since it's a key thing for this particular analytic

PR #517, opened 2024-03-14
Add push action handler logging

mmaatttt:
Also adding a basic search bar to the debug log

Edit: and now that I've figured out we can mock `UNNotificationResponse` in tests, I've updated `PushMonitor.didReceiveNotification(response: UNNotificationResponse, completionHandler: @escaping () -> Void) -> Bool` to take the full response object instead of just the `userInfo`. We'll need it for the `response.actionIdentifier` and other stuff in the future.

Simulator Screenshot - iPhone 15 Pro - 2024-03-14 at 11 03 48

### DIFF
--- a/Sources/AppcuesKit/Push/PushMonitor.swift
+++ b/Sources/AppcuesKit/Push/PushMonitor.swift
@@ -16,8 +16,7 @@ internal protocol PushMonitoring: AnyObject {
 
     func refreshPushStatus(completion: ((UNAuthorizationStatus) -> Void)?)
 
-    // Using `userInfo` as a parameter to be able to mock notification data.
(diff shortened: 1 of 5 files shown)


Sources/AppcuesKit/Push/PushMonitor.swift
```diff
@@ -17,12 +17,13 @@ internal protocol PushMonitoring: AnyObject {
     func refreshPushStatus(completion: ((UNAuthorizationStatus) -> Void)?)
 
     // Using `userInfo` as a parameter to be able to mock notification data.
```
iujames, line 19: can update the comment

iujames (approved):
nice!

PR #518, opened 2024-03-15
Send device_updated event on push status change

mmaatttt:
Per [discussion](https://app.shortcut.com/appcues/story/62368/implement-appcues-device-updated-event#activity-63884), when the app is foregrounded, the push status refresh should send a `appcues:device_updated` event if the status changed.

I've moved the `appcues:device_updated` event out of the request push action and now it can be triggered with the `publishChange` argument.

### DIFF
--- a/Sources/AppcuesKit/Presentation/Actions/Appcues/AppcuesRequestPushAction.swift
+++ b/Sources/AppcuesKit/Presentation/Actions/Appcues/AppcuesRequestPushAction.swift
@@ -32,14 +32,7 @@ internal class AppcuesRequestPushAction: AppcuesExperienceAction {
             }
 
             let pushMonitor = appcues.container.resolve(PushMonitoring.self)
-            let analyticsPublisher = appcues.container.resolve(AnalyticsPublishing.self)
-
-            pushMonitor.refreshPushStatus { _ in
-                analyticsPublisher.publish(TrackingUpdate(
-                    type: .event(name: Events.Device.deviceUpdated.rawValue, interactive: false),
-                    isInternal: true
-                ))
-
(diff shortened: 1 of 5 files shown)


Sources/AppcuesKit/Presentation/Actions/Appcues/AppcuesRequestPushAction.swift
```diff
@@ -33,13 +33,12 @@ internal class AppcuesRequestPushAction: AppcuesExperienceAction {
 
             let pushMonitor = appcues.container.resolve(PushMonitoring.self)
             let analyticsPublisher = appcues.container.resolve(AnalyticsPublishing.self)
+            analyticsPublisher.publish(TrackingUpdate(
+                type: .event(name: Events.Device.deviceUpdated.rawValue, interactive: false),
+                isInternal: true
+            ))
```
iujames, line 39: sorry im having trouble following the logic, assuming it works but does this trigger an update on the result handler, and then L41 would trigger another since the current status changed?

iujames (approved):

Sources/AppcuesKit/Presentation/Actions/Appcues/AppcuesRequestPushAction.swift
mmaatttt: This shouldn't be here at all, I must've inadvertently pasted it. Should just be this, which hopefully makes more sense. Thanks for catching this!

```swift
let pushMonitor = appcues.container.resolve(PushMonitoring.self)
pushMonitor.refreshPushStatus(publishChange: true) { _ in
    completion()
}
```

PR #520, opened 2024-03-18
Add support for push trigger deep links

mmaatttt:
Stacked on #519 (no real reason, but it's just nice to work with all the latest functionality)

### DIFF
--- a/Sources/AppcuesKit/Presentation/ExperienceLoader.swift
+++ b/Sources/AppcuesKit/Presentation/ExperienceLoader.swift
@@ -17,6 +17,13 @@ internal protocol ExperienceLoading: AnyObject {
         trigger: ExperienceTrigger,
         completion: ((Result<Void, Error>) -> Void)?
     )
+
+    func loadPush(
+        id: String,
+        published: Bool,
+        queryItems: [URLQueryItem],
+        completion: ((Result<Void, Error>) -> Void)?
+    )
 }
 
 @available(iOS 13.0, *)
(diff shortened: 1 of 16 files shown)


Sources/AppcuesKit/Presentation/ExperienceLoader.swift
```diff
@@ -17,6 +17,13 @@ internal protocol ExperienceLoading: AnyObject {
         trigger: ExperienceTrigger,
         completion: ((Result<Void, Error>) -> Void)?
     )
+
+    func loadPush(
+        id: String,
+        published: Bool,
+        queryItems: [URLQueryItem],
+        completion: ((Result<Void, Error>) -> Void)?
+    )
```
iujames, line 26: slightly funky looking to have push loading in `ExperienceLoading`. It would also seem odd to have in `PushMonitoring`. Maybe some other small util under the `/Push` folder? But not terribly critical. Certainly a small addition here at this point.

iujames (approved):
checking understanding - there is prob no reason to worry about a delay or anything, as the push can come in, foreground, and then the user can decide if they want to close the app, view from notification center, cold launch the app from it -- all still possible through other interactions.

Sources/AppcuesKit/Presentation/ExperienceLoader.swift
iujames: `ContentLoading` ? or leaving as-is is fine with me too, just came to mind looking it over.

mmaatttt: `ContentLoading` is good. I'll sleep on it and use that in the morning if nothing better comes to mind.

PR #521, opened 2024-03-19
Omit analytics from test/preview pushes

mmaatttt:


iujames (approved):
note: right now the sender is tracking sent/failed/skipped analytics on test, but including this test flag. I think that is ok and reporting can filter it out as needed. Seems useful to know if anything was actually processed by the BE, but also seems correct here to not send anything from the client - same as flow previews.

we can filter out the BE analytics too though in the future if we'd rather go that route and rely on other logs

PR #522, opened 2024-03-20
Determine push attachment file type using file mime type

mmaatttt:
Removing the requirement for `appcues_attachment_type`.

`URLSession.downloadTask(with:completionHandler:)` is almost as easy as `Data(contentsOf:)`. It creates a download task that saves the results to a file which we can copy to our attachment with the mime type of the `URLResponse`. Using `UTType` means we don't need a hardcoded map of mime types to file extensions (even if the pre-ios 14 fallback is a bit ugly).


iujames (approved):
nice! removed from docs and will remove from BE processing next, thanks.

PR #526, opened 2024-04-11
Add one line push config

mmaatttt:
This PR is very much building on the approach for the `UIScrollViewDelegate` swizzling added and explained in  https://github.com/appcues/appcues-ios-sdk/pull/479.

As noted in the `Appcues.swift` docs, there's 5 things that need to happen for automatic push config:
1. Calling `UIApplication.registerForRemoteNotifications()`.
    - This is straightforward.
2. Implementing `UIApplicationDelegate.application(_:didRegisterForRemoteNotificationsWithDeviceToken:)` to call `Appcues.setPushToken(_:)`.
    - Because we can safely assume there's always a `UIApplicationDelegate`, we just need to add the placeholder implementation if there is none, and then swizzle the method.
3. Ensuring `UNUserNotificationCenter.current().delegate` is set.
    - This is the same approach as the `UIScrollViewDelegate` getter swizzling.
4. Implementing `UNUserNotificationCenterDelegate.userNotificationCenter(_:didReceive:withCompletionHandler:)` to call `Appcues.didReceiveNotification(response:completionHandler:)`.
    - This is swizzled with a placeholder implementation from the getter of the `UNUserNotificationCenterDelegate` (same ideas as the `UIScrollView` approach.
5. Implementing `UNUserNotificationCenterDelegate.userNotificationCenter(_:willPresent:withCompletionHandler:)`.
    - This is swizzled with a placeholder implementation from the getter of the `UNUserNotificationCenterDelegate` (same ideas as the `UIScrollView` approach.

`AppcuesUNUserNotificationCenterDelegate` has a static instance that the various swizzled methods call. To properly support multiple SDK instances (that each can be auto configured!) there's a list of `pushMonitors` that we can operate from (using the weak wrapper approach from `AnalyticsPublisher`).

## Structural Notes
1. I moved the implementation of `Appcues.setPushToken` into `PushMonitor` so that the swizzled implementation can also easy access it.
2. Instead of having the almost the same long swizzling block for `UIScrollViewDelegate`, `UIApplicationDelegate`, and `UNUserNotificationCenterDelegate`, I created Swizzler with one extra parameter.

## What's Next
1. I've [thought about](https://appcues.slack.com/archives/C031RLGS4F8/p1712774012435019?thread_ts=1712773510.279779&cid=C031RLGS4F8) adding config options to control what specifically gets automatically swizzled, but for now I'm comfortable with the all-or-nothing approach.
2. I'd like tests for the swizzled behaviour (and the various scenarios of what's already implemented in the existing app), but I haven't yet looked into what's doable there.

### DIFF
--- a/Examples/DeveloperCocoapodsExample/CocoapodsExample/AppDelegate.swift
+++ b/Examples/DeveloperCocoapodsExample/CocoapodsExample/AppDelegate.swift
@@ -17,8 +17,12 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     ) -> Bool {
         // Override point for customization after application launch.
 
-        setupPush(application: application)
-        
+        // Automatically configure for push notifications
+        Appcues.shared.enableAutomaticPushConfig()
+
+        // Or, manually configure for push notifications
+        // setupPush(application: application)
+
         return true
     }
 
--- a/Examples/DeveloperCocoapodsExample/CocoapodsExample/AppDelegate.swift
+++ b/Examples/DeveloperCocoapodsExample/CocoapodsExample/AppDelegate.swift
@@ -18,7 +18,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         // Override point for customization after application launch.
 
         // Automatically configure for push notifications
-        Appcues.shared.enableAutomaticPushConfig()
+        Appcues.enableAutomaticPushConfig()
 
         // Or, manually configure for push notifications
         // setupPush(application: application)
--- a/Sources/AppcuesKit/Appcues.swift
+++ b/Sources/AppcuesKit/Appcues.swift
@@ -245,14 +245,8 @@ public class Appcues: NSObject {
     /// `application(_:didRegisterForRemoteNotificationsWithDeviceToken:)` function:
     @objc
     public func setPushToken(_ deviceToken: Data?) {
-        storage.pushToken = deviceToken?.map { String(format: "%02x", $0) }.joined()
-
-        if sessionID != nil {
-            analyticsPublisher.publish(TrackingUpdate(
-                type: .event(name: Events.Device.deviceUpdated.rawValue, interactive: false),
-                isInternal: true
-            ))
-        }
+        let pushMonitor = container.resolve(PushMonitoring.self)
+        pushMonitor.setPushToken(deviceToken)
     }
 
     /// Register a trait that modifies an `Experience`.
@@ -316,6 +310,25 @@ public class Appcues: NSObject {
         // resolving will init UIKitScreenTracking, which sets up the swizzling of
         // UIViewController for automatic screen tracking
         _ = container.resolve(UIKitScreenTracker.self)
+    }
+
+    /// Enables automatic push notification management.
+    ///
+    /// This should be called in `UIApplicationDelegate.application(_:didFinishLaunchingWithOptions:)` to ensure no incoming notifications are missed.
+    ///
+    /// The following will automatically be handled:
+    /// 1. Calling `UIApplication.registerForRemoteNotifications()`
+    /// 2. Implementing `UIApplicationDelegate.application(_:didRegisterForRemoteNotificationsWithDeviceToken:)`
+    /// to call ``setPushToken(_:)``
+    /// 3. Ensuring `UNUserNotificationCenter.current().delegate` is set
+    /// 4. Implementing `UNUserNotificationCenterDelegate.userNotificationCenter(_:didReceive:withCompletionHandler:)`
+    /// to call ``didReceiveNotification(response:completionHandler:)``
+    /// 5. Implementing `UNUserNotificationCenterDelegate.userNotificationCenter(_:willPresent:withCompletionHandler:)`
+    /// to show notification while the app is in the foreground
+    @objc
+    public func enableAutomaticPushConfig() {
+        let pushMonitor = container.resolve(PushMonitoring.self)
+        pushMonitor.configureAutomatically()
     }
 
     /// Verifies if an incoming URL is intended for the Appcues SDK.
--- a/Sources/AppcuesKit/Appcues.swift
+++ b/Sources/AppcuesKit/Appcues.swift
@@ -99,6 +99,25 @@ public class Appcues: NSObject {
         config.logger.info("Appcues SDK %{public}@ initialized", version())
     }
 
+    /// Enables automatic push notification management.
+    ///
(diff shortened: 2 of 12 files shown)


Sources/AppcuesKit/Push/UNUserNotificationCenter+AutoConfig.swift
```diff
@@ -0,0 +1,180 @@
...
+            swizzleSelector: #selector(appcues__userNotificationCenterDidReceive)
+        )
+
+        Swizzler.swizzle(
+            targetInstance: delegate,
+            targetSelector: NSSelectorFromString("userNotificationCenter:willPresentNotification:withCompletionHandler:"),
+            replacementOwner: UNUserNotificationCenter.self,
+            placeholderSelector: #selector(appcues__placeholderUserNotificationCenterWillPresent),
+            swizzleSelector: #selector(appcues__userNotificationCenterWillPresent)
+        )
```
mmaatttt, line 118: @iujames the [scrollview implementation](https://github.com/appcues/appcues-ios-sdk/blob/96dd818342a5eff6ef7099e2c843cc113a6dff21/Sources/AppcuesKit/Presentation/Extensions/UIScrollView%2BScrollObserver.swift#L112-L131) did this where the methods are swizzled every time the getter is called.

This seems like overkill, and maybe the methods could be swizzled once after the delegate getter is swizzled? Not sure if you have any insight here, or I can just try it and see what happens.

Sources/AppcuesKit/Appcues.swift
```diff
@@ -318,6 +312,25 @@ public class Appcues: NSObject {
...
+    /// 1. Calling `UIApplication.registerForRemoteNotifications()`
+    /// 2. Implementing `UIApplicationDelegate.application(_:didRegisterForRemoteNotificationsWithDeviceToken:)`
+    /// to call ``setPushToken(_:)``
+    /// 3. Ensuring `UNUserNotificationCenter.current().delegate` is set
+    /// 4. Implementing `UNUserNotificationCenterDelegate.userNotificationCenter(_:didReceive:withCompletionHandler:)`
+    /// to call ``didReceiveNotification(response:completionHandler:)``
+    /// 5. Implementing `UNUserNotificationCenterDelegate.userNotificationCenter(_:willPresent:withCompletionHandler:)`
+    /// to show notification while the app is in the foreground
+    @objc
+    public func enableAutomaticPushConfig() {
```
mmaatttt, line 329: Note that this is a function on `Appcues` instead of an `Appcues.Config` option because,

1. It’s easier to explain in the documentation that this should be called from `AppDelegate.didFinishLaunchingWithOptions` otherwise “assigning a delegate after the system calls these methods might cause you to miss incoming notifications.” A config option would mean being more prescriptive about how the `Appcues` instance gets inited.
2. It matches `Appcues.trackScreens()` above.

Examples/DeveloperCocoapodsExample/CocoapodsExample/AppDelegate.swift
```diff
@@ -17,8 +17,12 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     ) -> Bool {
         // Override point for customization after application launch.
 
-        setupPush(application: application)
-        
+        // Automatically configure for push notifications
+        Appcues.shared.enableAutomaticPushConfig()
+
+        // Or, manually configure for push notifications
+        // setupPush(application: application)
```
iujames, line 24: nice to still have the old configuration reference here in the example 👍

Sources/AppcuesKit/Push/PushMonitor.swift
```diff
@@ -65,6 +73,25 @@ internal class PushMonitor: PushMonitoring {
         refreshPushStatus(publishChange: true)
     }
 
+    func configureAutomatically() {
```
iujames, line 76: should this function set `configuredAutomatically` true?

Sources/AppcuesKit/Push/UNUserNotificationCenter+AutoConfig.swift
```diff
@@ -0,0 +1,180 @@
...
+
+        // If we need to set a non-nil implementation where there previously was not one,
+        // swap the swizzled getter back first, then assign, then restore the swizzled getter.
+        // This is done to avoid infinite recursion in some cases observed, where a UICollectionView,
```
iujames, line 122: this part _may_ not be needed here, was a weird issue with UICollectionView if I recall, but also probably not hurting anything.

Sources/AppcuesKit/Push/UNUserNotificationCenter+AutoConfig.swift
iujames: yeah it does seem like it shouldn't be needed each time, but I'm not 100% sure. I think it was challenging to know where to keep that state perhaps? but maybe some associated object type of state reference would work here?

iujames (approved):
looks great! some complicated stuff swizzling delegates, but this seems like it will be extremely handy to make the default case work quickly for customer apps.

Sources/AppcuesKit/Push/PushMonitor.swift
mmaatttt: Indeed it should!

Sources/AppcuesKit/Push/UNUserNotificationCenter+AutoConfig.swift
mmaatttt: Right, so if we remove `self.delegate = delegate` that's ok because the getter always ends up in the else here on each get?

```swift
if let existingDelegate = appcues__getNotificationCenterDelegate() {
    delegate = existingDelegate
} else {
    // if it is nil, then we assign our own delegate implementation so there is
    // something hooked in to listen to scroll
    delegate = AppcuesUNUserNotificationCenterDelegate.shared
}
```

Or we could keep `self.delegate = delegate` (just without the double swizzle bracket) which would mean the first time we'd hit the `else` above, but all subsequent times would hit the `if` where `existingDelegate == AppcuesUNUserNotificationCenterDelegate.shared`.

iujames: yeah I think you can move `self.delegate = delegate` right here in the `else` block, after it is created the first time. If I recall, thats what I had done initially, then ran into the other edge case with the scrolling implementation.

mmaatttt: Ok, I've updated to do that 👍

PR #528, opened 2024-04-16
Update push auto config to be static

mmaatttt:
This is a tweak targeting #526 in its own PR for readability. I've made the swizzling automatic push config static on `Appcues` for a couple reasons:

1. Swizzling affects the whole app, not just the instance. So in the theoretical multiple Appcues instance app, it'd be very strange to suggest that you could automatically configure one instance but not the other.
2. A static config option better sets us up for x-plat success. More [here](https://appcues.slack.com/archives/C031RLGS4F8/p1712932756018019), but since the x-plat libraries create the `Appcues` instance on-demand, that may be too late for the config necessary in `didFinishLaunching`.

The static observer list pattern I've added means that,
1. If an `Appcues` instance is created before the push auto config, then it's already in the observer list for when the swizzled methods start getting called, and
2. If the auto config is done, and then a new `Appcues` instant is created, the instance will add itself to the observer list and start receiving any push events from that point forward.

### DIFF
--- a/Sources/AppcuesKit/Push/UIApplication+AutoConfig.swift
+++ b/Sources/AppcuesKit/Push/UIApplication+AutoConfig.swift
@@ -35,10 +35,12 @@ extension UIApplication {
         _ application: UIApplication,
         didRegisterForRemoteNotificationsWithDeviceToken deviceToken: Data
     ) {
-        AppcuesUNUserNotificationCenterDelegate.shared.didRegister(deviceToken: deviceToken)
+        PushAutoConfig.didRegister(deviceToken: deviceToken)
 
         // Also call the original implementation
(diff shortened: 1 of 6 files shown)


Sources/AppcuesKit/Push/UIApplication+AutoConfig.swift
```diff
@@ -35,10 +35,12 @@ extension UIApplication {
         _ application: UIApplication,
         didRegisterForRemoteNotificationsWithDeviceToken deviceToken: Data
     ) {
-        AppcuesUNUserNotificationCenterDelegate.shared.didRegister(deviceToken: deviceToken)
```
mmaatttt, line 38: Having the new `PushAutoConfig` is nice because it means there's no out of place reference to `AppcuesUNUserNotificationCenterDelegate` any more.

iujames (approved):

PR #529, opened 2024-04-16
Add new push analytics properties

mmaatttt:


Sources/AppcuesKit/Push/ParsedNotification.swift
```diff
@@ -34,7 +36,9 @@ internal struct ParsedNotification {
         self.userID = userID
         self.notificationID = notificationID
 
+        self.notificationVersion = userInfo["appcues_notification_version"] as? Int
```
iujames, line 39: these are long timestamp values from the BE, double checking if Int will handle that or we need Int64? I don't recall, too many language context switches lately. I believe it is same as `Experience.publishedAt` which is Int, so maybe this is good.

iujames (approved):

Sources/AppcuesKit/Push/ParsedNotification.swift
mmaatttt: > On 32-bit platforms, Int is the same size as Int32, and on 64-bit platforms, Int is the same size as Int64.

https://developer.apple.com/documentation/swift/int

PR #530, opened 2024-04-17
Update push auto config to store values for future instances

mmaatttt:
This PR adds automatic push config support for `Appcues` instances created after `Appcues.enableAutomaticPushConfig()` is called. `enableAutomaticPushConfig()` must be called from `didFinishLaunchingWithOptions`, but the `Appcues` instance might not be inited until after (especially in the case of Democues!)

There's 2 things to handle:
1. The push token. We save the most recent value and give it to the new instance when it's registered.
2. A notification response. If the notification response was handled by an existing instance, we can ignore it, but otherwise save the most recent one for a new instance to try and handle.

The end result is that in Democues, a notification can (cold) launch the app, and once a matching account ID and app ID are entered (which creates the `Appcues` instance), the push token will be saved and the notification handler will execute seamlessly:

https://github.com/appcues/appcues-ios-sdk/assets/845681/e233348b-f53e-4644-8698-a8facb384216

### DIFF
--- a/Sources/AppcuesKit/Push/PushAutoConfig.swift
+++ b/Sources/AppcuesKit/Push/PushAutoConfig.swift
@@ -12,8 +12,26 @@ internal enum PushAutoConfig {
     // This is an array to support the (rare) case of multiple SDK instances supporting push
     private static var pushMonitors: [WeakPushMonitoring] = []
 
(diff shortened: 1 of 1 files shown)


iujames (approved):
beautifully flexible 🎉

PR #537, opened 2024-05-07
Add push preview error feedback toast

iujames:
Gives some immediate feedback if something went wrong on the push preview request, similar to flow preview issues.
Screenshot 2024-05-07 at 12 56 08 PM

Also, now provides error detail in the detailed log with the response body error messages to troubleshoot. (note: some work going in api/pushserv soon to improve a few cases here to be more helpful)
Screenshot 2024-05-07 at 12 56 28 PM

### DIFF
--- a/Sources/AppcuesKit/Data/Networking/NetworkClient.swift
+++ b/Sources/AppcuesKit/Data/Networking/NetworkClient.swift
@@ -173,11 +173,12 @@ internal class NetworkClient: Networking {
         _ urlRequest: URLRequest,
         completion: @escaping (_ result: Result<Void, Error>) -> Void
     ) {
-        let dataTask = config.urlSession.dataTask(with: urlRequest) { [weak self] _, response, error in
+        let dataTask = config.urlSession.dataTask(with: urlRequest) { [weak self] data, response, error in
             let url = (response?.url ?? urlRequest.url)?.absoluteString ?? "<unknown>"
             let statusCode = (response as? HTTPURLResponse)?.statusCode ?? -1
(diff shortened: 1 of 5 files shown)


Sources/AppcuesKit/Data/Networking/NetworkClient.swift
```diff
@@ -173,11 +173,12 @@ internal class NetworkClient: Networking {
...
         completion: @escaping (_ result: Result<Void, Error>) -> Void
     ) {
-        let dataTask = config.urlSession.dataTask(with: urlRequest) { [weak self] _, response, error in
+        let dataTask = config.urlSession.dataTask(with: urlRequest) { [weak self] data, response, error in
             let url = (response?.url ?? urlRequest.url)?.absoluteString ?? "<unknown>"
             let statusCode = (response as? HTTPURLResponse)?.statusCode ?? -1
+            let logData = String(data: error?.data ?? data ?? Data.empty, encoding: .utf8) ?? ""
 
-            self?.config.logger.debug("RESPONSE: %{public}d %{public}@", statusCode, url)
+            self?.config.logger.debug("RESPONSE: %{public}d %{public}@\n%{private}@", statusCode, url, logData)
```
iujames, line 181: @mmaatttt curious your thoughts on this part - in the function above this, we log the response/error body, but in this one we did not previously. I think it's helpful to have in either case here, granted this version does not actually deserialize the response into anything sent back to caller.

mmaatttt: Agreed. Probably just an oversight.

mmaatttt (approved):

PR #541, opened 2024-05-22
Add support for automatic push environment detection

mmaatttt:
Add a `_pushEnvironment` device auto property with a value of `development` or `production`. The value starts as unknown (which tracks as `production`) until a background thread can read and parse the provisioning profile data, if it's available.

There's a few options here:

1. No `embedded.mobileprovision`. This could be a simulator build ( so want `development`) or an App Store one (`production`):
    > App Store apps do not contain an embedded provisioning profile because the App Store checks that the app is signed and provisioned correctly as part of its app ingestion process.
      - https://developer.apple.com/documentation/technotes/tn3125-inside-code-signing-provisioning-profiles#Profile-location
2. There is an `embedded.mobileprovision`. This could a development build (`development`) or an Ad Hoc build (`production`). In this case we can deserialize the provisioning profile and determine the correct environment, falling back to `production` if anything is unexpected. There's [a](https://www.process-one.net/blog/reading-ios-provisioning-profile-in-swift/) [few](https://gist.github.com/mremond/7c62e85e2f1ee6ae311ee039b959fcc0) [examples](https://stackoverflow.com/a/69905152) of doing this out there.

I haven't been able to test most of the error cases in a meaningful way because I can't modify the `embedded.mobileprovision` to do something unexpected. From what I've observed, most of the errors should never happen, but at least we'll be able to narrow things down if they do.

I have tested with a simulator, development build, and TestFlight (App Store) build. Our alpha/beta period will provide more opportunities to test different configuration.

### DIFF
--- a/Sources/AppcuesKit/Push/PushEnvironment.swift
+++ b/Sources/AppcuesKit/Push/PushEnvironment.swift
@@ -1,0 +1,144 @@
+//
+//  PushEnvironment.swift
+//  Appcues
+//
+//  Created by Matt on 2024-05-21.
+//
+
+import UIKit
+
+internal enum PushEnvironment {
+    case unknown(Reason)
+    case development
+    case production
+
+    enum PushEnvironmentError: LocalizedError {
+        case noEntitlementsKey
+        case noEnvironmentKey
+        case unexpectedEnvironment(String)
+
+        var errorDescription: String? {
+            switch self {
+            case .noEntitlementsKey: return "No 'Entitlements' key"
+            case .noEnvironmentKey: return "No entitlement 'aps-environment' key"
+            case .unexpectedEnvironment(let value): return "Unexpected 'aps-environment' value '\(value)'"
+            }
+        }
+    }
+
+    enum Reason {
+        case notComputed, error(Error)
+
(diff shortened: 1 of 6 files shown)


Sources/AppcuesKit/Push/PushEnvironment.swift
```diff
@@ -0,0 +1,144 @@
...
+
+        guard let pushEnvironment = PushEnvironment(value: environment) else {
+            return .unknown(.error(PushEnvironment.PushEnvironmentError.unexpectedEnvironment(environment)))
+        }
+
+        return pushEnvironment
+    } catch ProvisioningProfileError.noEmbeddedProfile {
+        // App Store apps do not contain an embedded provisioning profile,
+        // and since we know we're not on a simulator, that means it's "production".
+        return .production
```
mmaatttt, line 96: This is the important part to not show a debugger error for App Store builds ([slack ref](https://appcues.slack.com/archives/C031RLGS4F8/p1716391502737889?thread_ts=1716296544.624109&cid=C031RLGS4F8))

iujames (approved):
awesome!

PR #542, opened 2024-05-27
Show errors from push test endpoint

mmaatttt:
I needed a way to decode the data from the error path, so now NetworkingError.nonSuccessfulStatusCode holds the error response data to decode in `PushVerifier`.


Sources/AppcuesKit/Presentation/Debugger/PushVerifier.swift
```diff
@@ -64,8 +64,8 @@ internal class PushVerifier {
                 return "Error 9: Foreground presentation handler not implemented"
             case .noForegroundPresentationOption:
                 return "Note: Application is not configured to display foreground notifications"
-            case .serverError:
-                return "Error 500: Server Error"
+            case .serverError(let error):
+                return "Error 500: \(error ?? "Unexpected server error")"
```
iujames, line 68: can we use the status code from the error instead of 500? it could be different things like 400 or 404 in some of these cases I believe.

iujames (approved):

Sources/AppcuesKit/Presentation/Debugger/PushVerifier.swift
mmaatttt: oh good idea!

PR #545, opened 2024-06-18
Ensure latest device prop values are always sent

mmaatttt:
James [noted](https://appcues.slack.com/archives/C031RLGS4F8/p1717616763921679) a while back that there can be a race where `pushEnabled` is read and sent to the backend at startup before the correct value can be determined.

To fix this, we want to make sure the initial request for the permission status will send a device updated event if the value is different than the one that was used. This initial request was the only place with a false argument like `refreshPushStatus(publishChange: false)`.

So I've removed the `publishChange` parameter in favour of calculating a `shouldPublish` property that confirms that the value has changed AND (new) that there's an active session (this means that the previous value was used).

I've also added the same check to `setPushToken` so we're not triggering device update events with an unchanged push token value.

### DIFF
--- a/Sources/AppcuesKit/Push/PushMonitor.swift
+++ b/Sources/AppcuesKit/Push/PushMonitor.swift
@@ -17,7 +17,7 @@ internal protocol PushMonitoring: AnyObject {
 
     func setPushToken(_ deviceToken: Data?)
 
-    func refreshPushStatus(publishChange: Bool, completion: ((UNAuthorizationStatus) -> Void)?)
+    func refreshPushStatus(completion: ((UNAuthorizationStatus) -> Void)?)
 
     func didReceiveNotification(response: UNNotificationResponse, completionHandler: @escaping () -> Void) -> Bool
     @discardableResult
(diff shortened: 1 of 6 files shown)


Sources/AppcuesKit/Push/PushMonitor.swift
```diff
@@ -94,7 +96,7 @@ internal class PushMonitor: PushMonitoring {
         #endif
 
         UNUserNotificationCenter.current().getNotificationSettings { [weak self] settings in
-            let shouldPublish = publishChange && self?.pushAuthorizationStatus != settings.authorizationStatus
+            let shouldPublish = self?.appcues?.sessionID != nil && self?.pushAuthorizationStatus != settings.authorizationStatus
```
andretortolano, line 99: can you expand on this change? is that to ensure you are not starting a session on this event ?

andretortolano (approved):

Sources/AppcuesKit/Push/PushMonitor.swift
mmaatttt: Basically yeah. There's [the code](https://github.com/appcues/appcues-ios-sdk/blob/5b3668a18154dcfc37a5b551658c29985a5f80e1/Sources/AppcuesKit/Data/Analytics/AnalyticsPublisher.swift#L55-L56) that starts a session if it can (ie there's a stored user id) and we don't want to do that in this case.

If there's a session and the value is different, that means that the DB has the old (now wrong) value and we want to update it.

PR #548, opened 2024-07-03
Ensure push completion handlers are always called

mmaatttt:
Was working on push docs and noticed a two gaps in how the automatic push configuration calls completion handlers. For context:

[Docs](https://developer.apple.com/documentation/usernotifications/unusernotificationcenterdelegate/usernotificationcenter(_:didreceive:withcompletionhandler:)) for `userNotificationCenter(_:didReceive:withCompletionHandler:)`:
> You must execute this block at some point after processing the user’s response to let the system know that you are done.

Docs for `userNotificationCenter(_:willPresent:withCompletionHandler:`:
> Always execute this block at some point during your implementation of this method.

First off, in `PushAutoConfig`, the completion handler wasn't being called in the case where there's not an Appcues instance to immediately handle the push, but we save it for deferred handling. I've added the call and some inline docs to make it clearer what's happening.

Then, for the cases where an incoming notification isn't an Appcues one and the app hasn't already implemented the methods, the swizzling implementation [falls back](https://github.com/appcues/appcues-ios-sdk/blob/9e561d758b9f397ed1d09e9bf7bb6e150dae4382/Sources/AppcuesKit/Utilities/Swizzler.swift#L42-L57) ([here](https://github.com/appcues/appcues-ios-sdk/blob/9e561d758b9f397ed1d09e9bf7bb6e150dae4382/Sources/AppcuesKit/Push/UNUserNotificationCenter%2BAutoConfig.swift#L96-L97) and [here](https://github.com/appcues/appcues-ios-sdk/blob/9e561d758b9f397ed1d09e9bf7bb6e150dae4382/Sources/AppcuesKit/Push/UNUserNotificationCenter%2BAutoConfig.swift#L110-L111)) to an "empty" default implementation that we provide. So those default implementations need to call the completion block as well.

### DIFF
--- a/Sources/AppcuesKit/Push/PushAutoConfig.swift
+++ b/Sources/AppcuesKit/Push/PushAutoConfig.swift
@@ -66,12 +66,20 @@ internal enum PushAutoConfig {
         let didHandleResponse = pushMonitors.contains { weakPushMonitor in
             guard let pushMonitor = weakPushMonitor.value else { return false }
 
+            // `completionHandler` is executed iff this returns true
             return pushMonitor.didReceiveNotification(response: response, completionHandler: completionHandler)
         }
 
-        // Save unhandled response for any future PushMonitor instances
-        // (or clear the existing value if a newer response was handled)
-        receivedUnhandledResponse = !didHandleResponse ? response : nil
+        if didHandleResponse {
+            // Clear the existing value if a newer response was handled
+            receivedUnhandledResponse = nil
+        } else {
+            // Save unhandled response for any future PushMonitor instances
+            receivedUnhandledResponse = response
+            // Completion block has yet to be called because no PushMonitor instance has handled it
+            // The block must be called by all paths of this function, so call it now
+            completionHandler()
+        }
     }
 
     // Shared instance is called from the swizzled method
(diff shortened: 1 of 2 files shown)


Sources/AppcuesKit/Push/PushAutoConfig.swift
```diff
@@ -66,12 +66,20 @@ internal enum PushAutoConfig {
         let didHandleResponse = pushMonitors.contains { weakPushMonitor in
             guard let pushMonitor = weakPushMonitor.value else { return false }
 
+            // `completionHandler` is executed iff this returns true
```
andretortolano, line 69: typo in comment (iff?)

mmaatttt: Actually intentional! _iff_ ie "[if and only if](https://en.wikipedia.org/wiki/If_and_only_if)"

andretortolano (approved):

Sources/AppcuesKit/Push/PushAutoConfig.swift
iujames: should it be `iff this returns false` ? Or am I misreading that?

mmaatttt: The closure block passed into `pushMonitor.didReceiveNotification` is called only when it returns true. Otherwise it's uncalled. Because this line is the return value in the `didHandleResponse = pushMonitors.contains` block, `didHandleResponse` then indicates if the completion handler has been called.

I definitely need to make this comment clearer, so I'll do that in a future PR 😅

iujames: ah ok I follow now, thanks!

PR #564, opened 2024-09-30
Add support for swipe-to-dismiss of slideouts

mmaatttt:
The idea here is to allow slideouts to swipe-to-dismiss similar to how a cover/half sheet would. That means a few things:
1. Only allowed when `isModalInPresentation == false`. This is controlled by the skippable trait. Otherwise the view can be dragged a bit, but rubber-bands back to the original position.
2. The view can be dragged in the dismissal direction (down for a sheet, to an adjacent edge for a slideout). Dragging otherwise (up for a sheet, towards non-adjacent edges for a slideout) will respond, but with a rubber-band effect to indicate it's being dragged in an unsupported direction.
3. The drag gestures are appropriately responsive, smooth, and feel "right".

Nothing here is too complicated. Process is as follows:
1. On presenting the modal, figure out which edges are allowed for drag-to-dismiss (`DragDirection.allowedEdges`) and attach the pan gesture
2. On pan, move the view around, scaling back the distance if dragging in a disallowed direction (`DragDirection.newPoint`)
3. On pan end, figure out if the view would end up outside the screen bounds and and dismiss the view controller if it will (`DragDirection.isSufficientToDismiss`)

The actual gesture handling is quite similar to the debugger floating button and uses the techniques from the classic [Building Fluid Interfaces](https://medium.com/@nathangitter/building-fluid-interfaces-ios-swift-9732bb934bf5) article.

### DIFF
--- a/Sources/AppcuesKit/Presentation/Traits/Appcues/ExperienceWrapperViewController+DragDirection.swift
+++ b/Sources/AppcuesKit/Presentation/Traits/Appcues/ExperienceWrapperViewController+DragDirection.swift
@@ -1,0 +1,91 @@
+//
+//  ExperienceWrapperViewController+DragDirection.swift
+//  AppcuesKit
+//
+//  Created by Matt on 2024-09-27.
+//  Copyright © 2024 Appcues. All rights reserved.
+//
+
+import SwiftUI
+
+@available(iOS 13.0, *)
+extension ExperienceWrapperViewController {
+    struct DragDirection {
+        enum Edge {
+            case left, right, up, down
+        }
+
+        let xDirection: Edge
+        let yDirection: Edge
+
+        private let initialPoint: CGPoint
+        private let newPoint: CGPoint
+
+        init(initial: CGPoint, new: CGPoint) {
+            self.initialPoint = initial
+            self.newPoint = new
+            self.xDirection = new.x > initial.x ? .right : .left
+            self.yDirection = new.y > initial.y ? .down : .up
+        }
+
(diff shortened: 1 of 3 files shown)


Sources/AppcuesKit/Presentation/Traits/Appcues/ExperienceWrapperViewController+DragDirection.swift
```diff
@@ -0,0 +1,91 @@
...
+
+        init(initial: CGPoint, new: CGPoint) {
+            self.initialPoint = initial
+            self.newPoint = new
+            self.xDirection = new.x > initial.x ? .right : .left
+            self.yDirection = new.y > initial.y ? .down : .up
+        }
+
+        /// Determine which edges are eligible for drag-to-dismiss.
+        static func allowedEdges(horizontalAlignment: HorizontalAlignment, verticalAlignment: VerticalAlignment) -> Set<Edge> {
```
iujames, line 32: is this a Set for future extensibility to allow dragging away to multiple edges?

mmaatttt: Hah kinda... It's a set because I started off allowing dragging to multiple edges and then tweaked it to be restricted to a single edge since that felt better. For slideouts in the corners, ie bottom left, it's not the worst to allow it to be swiped to the left or the bottom, but my personal preference is only to the left

iujames (approved):

PR #566, opened 2024-10-02
Update modal context traits to present in their own UIWindow

mmaatttt:
The `@appcues/modal` and `@appcues/tooltip` trait now present via the new `ModalContextManager` which manages the lifecycle of a `UIWindow` that's overlaid over the application to present modals and tooltips. This PR sets the foundation for the backdrop passthrough logic that's up next.

## Notes
1. I created `ModalContextManager` to avoid having the exact same code in both the modal and tooltip traits. 
2. I considered having `ModalContextManager` only create a single window for the whole `Appcues` instance, but then ownership gets complicated and it's the job of the state machine anyways to make sure there's not multiple modals being presented.
3. There's an exception `useSameWindow: false` for the sheet modals that uses the old approach to preserve the scale animation of the presenting view controller.
4. `AppcuesUIWindow` needs an empty root view controller (instead of using the modal controller) to be able to do the expected presentation animation of the modal controller.
5. `AppcuesUIWindow` is similar to the `DebugUIWindow` in how it ignores hit events.

### DIFF
--- a/Sources/AppcuesKit/Presentation/ExperienceRendering/ModalContextManager.swift
+++ b/Sources/AppcuesKit/Presentation/ExperienceRendering/ModalContextManager.swift
@@ -1,0 +1,95 @@
+//
+//  ModalContextManager.swift
+//  AppcuesKit
+//
+//  Created by Matt on 2024-09-25.
+//  Copyright © 2024 Appcues. All rights reserved.
+//
+
+import UIKit
+
+@available(iOS 13.0, *)
+internal class ModalContextManager {
+    var presentingWindow: AppcuesUIWindow?
+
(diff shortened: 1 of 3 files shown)


Sources/AppcuesKit/Presentation/ExperienceRendering/ModalContextManager.swift
```diff
@@ -0,0 +1,95 @@
...
+//
+//  Created by Matt on 2024-09-25.
+//  Copyright © 2024 Appcues. All rights reserved.
+//
+
+import UIKit
+
+@available(iOS 13.0, *)
+internal class ModalContextManager {
+    var presentingWindow: AppcuesUIWindow?
```
iujames, line 13: do we need to hold this reference?

iujames (approved):
nice improvement!

PR #567, opened 2024-10-04
Ignore touches on backdrop with no traits

mmaatttt:
Here's the idea:
1. The `ExperienceWrapperView` becomes a `HitTestingOverrideUIView(overrideApproach: .ignoreSelf)` so only hits in it's content are captured
2. The `ExperienceWrapperView.backdropView` also becomes a `HitTestingOverrideUIView(overrideApproach: .ignoreSelf)` so it doesn't capture any hits unless it has subviews that do.
3. We update the `AppcuesBackdropDecoratingTrait` to make sure they use subviews that do/don't capture hits as desired
    1. `@appcues/backdrop` adds a subview now (instead of setting the color directly) that **does** capture hits
    2. `@appcues/effects` does `effectView.isUserInteractionEnabled` so it **does not** capture hits (note that we can still have effects without a backdrop and in that case the user can interact with the app under the confetti)
    3. `@appcues/target-element` and `@appcues/target-rectangle` use a `FrameObserverView` in the backdrop which now has `isUserInteractionEnabled = false` so it **does not** capture hits
    4. `@appcues/target-interaction` now checks for `metadata["backdropBackgroundColor"] != nil` before adding the target view to ensure that there's no target capture view if there's no backdrop (because it's weird to have an invisible section of the screen do something different)
    5. `@appcues/skippable` is unchanged, but it adds a gesture recognizer to `backdropView` so the skip backdrop tap only applies if the backdrop view is capturing a hit from one of the other traits

To summarize, touches will flow through to the app if there is no `@appcues/backdrop` trait 😎 

# Other approaches
My first attempt at this set `backdropView.alpha = 0` to disable hits and then had the backdrop trait set the alpha back to 1, but this was very fragile because other traits theoretically might need too change that value too and then slideout animator would change it and lots of things could go wrong, even if I could make it work right now. The backdrop subview approach is much nicer, and actually is what we recommend in the trait docs, so that;'s a win for forward thinking:
```
/// Prefer adding a subview to `backdropView` over modifying `backdropView` itself where possible
/// to preserve composability of multiple traits.
```

I also included the slideout swipe-to-dismiss update to not allow swiping for centered slideouts in this PR

### DIFF
--- a/Examples/DeveloperCocoapodsExample/CocoapodsExample/Base.lproj/Main.storyboard
+++ b/Examples/DeveloperCocoapodsExample/CocoapodsExample/Base.lproj/Main.storyboard
@@ -722,7 +722,6 @@ With embeds, this paradigm changes, and a qualification response may contain zer
                                         <buttonConfiguration key="configuration" style="tinted" title="Trigger Event 3"/>
                                         <connections>
                                             <action selector="buttonThreeTapped:" destination="IV4-7k-Qev" eventType="touchUpInside" id="CgE-x1-GH2"/>
-                                            <action selector="buttonTwoTapped:" destination="IV4-7k-Qev" eventType="touchUpInside" id="Mrc-oW-IJA"/>
                                         </connections>
                                     </button>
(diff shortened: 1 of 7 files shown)


iujames (approved):
Looks great - just to summarize it all, effectively touches will flow through to the app if there is no `@appcues/backdrop` trait, is that accurate?

mmaatttt:
@iujames 
>Looks great - just to summarize it all, effectively touches will flow through to the app if there is no @appcues/backdrop trait, is that accurate?


Yep, exactly. Apparently I'm feeling verbose today 😉

PR #573, opened 2024-11-05
Custom components

mmaatttt:
## How it works

1. Register a custom component (a `UIViewController` that conforms to `AppcuesCustomComponentViewController`) statically with `Appcues.registerCustomComponent()`
2. Add a new experience JSON component type `customComponent` that takes an identifier and arbitrary config dictionary
3. The `AppcuesCustomComponent` SwiftUI view uses a `CustomComponentRepresentable` that wraps the view controller that was registered

The config decoding is entirely configurable from the client app and the provided `AppcuesExperienceActions` allows a developer to easily interact with the appcues context. A bunch of the Appcues actions were updated to have inits that `AppcuesExperienceActions` can easily use.

The debugger was updated with a Plugins section where custom components are listed, and if they add a `static var debugConfig: [String: Any]?` property, there's a detail screen that renders the component with the provided properties. To do this we:
1. Have a `DebuggableDecoder` that serializes the debug config and provides it to the controller to deserialize normally
2. Create a `CustomComponentModel`, feeding it the `DebuggableDecoder` to render the view 
3. Have a debug `ExperienceStepViewModel` that injects a `DebugAppcuesExperienceActions` to log the actions instead of actually execute them

There's an example `video` component implementation in the example apps that we'll reference in a tutorial doc (coming soon). I also need to add docs to some of the new public interface, but I'm opening the PR first to gets some eyes on it.

### DIFF
--- a/Examples/DeveloperCocoapodsExample/CocoapodsExample/AppDelegate.swift
+++ b/Examples/DeveloperCocoapodsExample/CocoapodsExample/AppDelegate.swift
@@ -23,6 +23,8 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         // Or, manually configure for push notifications
         // setupPush(application: application)
 
+        Appcues.registerCustomComponent(identifier: "liveStream", type: LiveStreamViewController.self)
+
         return true
     }
 
--- a/Sources/AppcuesKit/Presentation/Public/Plugins/AppcuesExperiencePluginConfiguration.swift
+++ b/Sources/AppcuesKit/Presentation/Public/Plugins/AppcuesExperiencePluginConfiguration.swift
@@ -10,7 +10,7 @@ import Foundation
 
 /// An object that decodes instances of a plugin configuration from an Experience JSON model.
 @objc
-internal class AppcuesExperiencePluginConfiguration: NSObject {
+public class AppcuesExperiencePluginConfiguration: NSObject {
 
     /// Context in which a plugin can be applied.
     @objc
@@ -43,7 +43,7 @@ internal class AppcuesExperiencePluginConfiguration: NSObject {
     /// Returns a value of the type you specify, decoded from a JSON object.
     /// - Parameter type: The type of the value to decode from the supplied plugin decoder.
     /// - Returns: A value of the specified type, if the decoder can parse the data.
-    internal func decode<T: Decodable>(_ type: T.Type) -> T? {
+    public func decode<T: Decodable>(_ type: T.Type) -> T? {
         return decoder.decode(type)
     }
 }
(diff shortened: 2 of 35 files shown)


Sources/AppcuesKit/Presentation/Public/Plugins/AppcuesExperiencePluginConfiguration.swift
```diff
@@ -10,7 +10,7 @@ import Foundation
 
 /// An object that decodes instances of a plugin configuration from an Experience JSON model.
 @objc
-internal class AppcuesExperiencePluginConfiguration: NSObject {
+public class AppcuesExperiencePluginConfiguration: NSObject {
```
iujames, line 13: cool that this thing just worked in the repurposed usage here

Sources/AppcuesKit/Presentation/Traits/TraitComposer.swift
```diff
@@ -16,12 +16,14 @@ internal protocol TraitComposing: AnyObject {
 @available(iOS 13.0, *)
 internal class TraitComposer: TraitComposing {
 
+    private weak var appcues: Appcues?
```
iujames, line 19: is this used?

iujames (approved):
looking great!

Sources/AppcuesKit/Presentation/Traits/TraitComposer.swift
mmaatttt: Yes, the reference is passed into the `ExperienceStepViewModel` on L95 because it's needed as part of the custom component init (for `AppcuesExperiencePluginConfiguration` and `AppcuesExperienceActions`).

Sources/AppcuesKit/Presentation/Public/Plugins/AppcuesExperiencePluginConfiguration.swift
mmaatttt: Note that I only made public the parts we need (`decode`). If we made traits/actions public again, the `level` and `renderContext` would probably need to be public too.

Sources/AppcuesKit/Presentation/Traits/TraitComposer.swift
iujames: doh, my bad, i misread the diff here with a collapsed section. thank you

Examples/DeveloperCocoapodsExample/CocoapodsExample/AppDelegate.swift
```diff
@@ -23,6 +23,8 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         // Or, manually configure for push notifications
         // setupPush(application: application)
 
+        Appcues.registerCustomComponent(identifier: "video", type: VideoViewController.self)
```
andretortolano, line 26: will rename Android's to match this.

mmaatttt:
@iujames @andretortolano I added docs in [b399548](https://github.com/appcues/appcues-ios-sdk/pull/573/commits/b399548a65a4f630ee6f536c72f7f2a11feca0e1) if you don't mind giving it a read through for typos/clarity.

PR #577, opened 2024-11-20
[SDK5] Update iOS deployment target to 13

mmaatttt:
I created an `sdk5` branch to hold the new major version changes.

First off is updating the deployment target from iOS 11 to iOS 13. Doing this separately to help reduce the diff of the async/await changes. It's mostly removing `@available(iOS 13.0, *)` checks and then updating a handful of `if #available(iOS 13.0, *)` conditions.


iujames (approved):
ahhh feels good

note to selves to update any docs when the time comes to release it all, like https://docs.appcues.com/mobile-installation-overview/installing-the-mobile-sdk#ios-21, maybe others?

PR #578, opened 2024-11-22
bugifx: problem with recursive calls

wiktor7245:
Hi, i had a problem with Expo app which uses that in  "@appcues/expo-config" and "@appcues/react-native" packages. I had a config where i use "@react-native-firebase" packages to handle notifications. App was crashing on start. This PR is fixing this issue


mmaatttt:
Hi @wiktor7245,

Thanks for identifying this issue and ever more thanks for contributing a solution! We tested and reproduced the issue as you described it and we realized we actually have a solution that we previously implemented when swizzling the `UIScrollViewDelegate`. Can you give the code in #579 a try and confirm that it fixes the issue as well? For the sake of consistency across the codebase our preference would be to use that solution instead of this one.

mmaatttt:
@wiktor7245 `@appcues/react-native@4.3.1` include the crash fix in #579.

wiktor7245:
Hi again! I checked your branch in the version 4.3.2 and it seems to work. So i guess, my PR can be closed

PR #581, opened 2024-12-04
Concurrency warning fixes

mmaatttt:
This is tackling some of the low hanging fruit of warnings on the road to Swift 6.

1. `scrollIndicatorInsets` was deprecated.
2. Using a `Timer` causes isolation warnings, so I replaced those with Tasks that sleep. Good article here: https://wadetregaskis.com/performing-a-delayed-and-or-repeating-operation-in-a-swift-actor/
3. Fixed the easy isolated global variable warnings by changing `static var` to static let` and adding `nonisolated(unsafe)` where it makes sense. There are still other instances of this warning that will be a bit more work to resolve.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -21,7 +21,7 @@ orbs:
 executors:
   xcode_15:
     macos:
-      xcode: 15.1.0
+      xcode: 16.1.0
     resource_class: macos.m1.medium.gen1
     working_directory: /Users/distiller/project
     shell: /bin/bash --login -o pipefail
--- a/Sources/AppcuesKit/Presentation/Debugger/FloatingView/DebugView.swift
+++ b/Sources/AppcuesKit/Presentation/Debugger/FloatingView/DebugView.swift
@@ -24,8 +24,8 @@ internal class DebugView: UIView {
 
     private var initialTouchOffsetFromCenter: CGPoint = .zero
 
-    private var dismissViewTimer: Timer?
-    private var canDismiss: Bool { dismissViewTimer?.isValid == false }
+    private var dismissViewTimer: Task<Void, Never>?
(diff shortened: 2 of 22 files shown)


Sources/AppcuesKit/Presentation/Debugger/FloatingView/DebugView.swift
```diff
@@ -24,8 +24,8 @@ internal class DebugView: UIView {
 
     private var initialTouchOffsetFromCenter: CGPoint = .zero
 
-    private var dismissViewTimer: Timer?
-    private var canDismiss: Bool { dismissViewTimer?.isValid == false }
+    private var dismissViewTimer: Task<Void, Never>?
```
iujames, line 27: should the variable be renamed?

iujames (approved):

Sources/AppcuesKit/Presentation/Debugger/FloatingView/DebugView.swift
mmaatttt: It's still functioning as a timer, so I don't mind the name since it signals the intended purpose 🤷‍♂️

iujames: sounds good 👍

.circleci/config.yml
```diff
@@ -21,7 +21,7 @@ orbs:
 executors:
   xcode_15:
     macos:
-      xcode: 15.1.0
+      xcode: 16.1.0
```
iujames, line 24: oh nice, i missed that the previous update didn't already do this

mmaatttt: Didn't think about it until `nonisolated(unsafe)` caused the build to fail 😅

PR #586, opened 2025-01-28
👌 strip hyphen from form labels

andretortolano:
Customer reported mismatching keys between Android and iOS when a form is submitted. iOS needs to remove `-` the Hyphen same way Android and Web does when massaging data to send as part of the step interaction.


Tests/AppcuesKitTests/Actions/AppcuesSubmitFormActionTests.swift
```diff
@@ -30,6 +30,21 @@ class AppcuesSubmitFormActionTests: XCTestCase {
 
     func testExecuteValidForm() throws {
         // Arrange
+        let formItem = ExperienceData.FormItem(model: ExperienceComponent.TextInputModel(
```
andretortolano, line 33: Need help with the test, if I just change the expectedFormItem then when we use it to compare late its going to mismatch, but I think if I create a new one like I did here I'm not sure the test is actually doing whats supposed to do (its passing but maybe wrong)


Ideally I wanted to test the asSlurg separately but its part of the ExperienceData file as a private extension so I probably cant do it

iujames: I think the core of the test is the part on L90 - checking the generated profile attribute name.  You can change the expectedFormItem to have a `-` in the label, and then check that this gets excluded as expected on L90 below in the call to `.verifyPropertiesMatch(updates[0].properties)` (the profile update)

to change the baseline of the test though you'll need to make the same change to the label here in the mocks https://github.com/appcues/appcues-ios-sdk/blob/main/Tests/AppcuesKitTests/Mocks.swift#L183

mmaatttt: It's a bit complex. The label value comes all the way from [here](https://github.com/appcues/appcues-ios-sdk/blob/7cc917c928e43547a81301b2d8edd9e1f59347c0/Tests/AppcuesKitTests/Mocks.swift#L183). 

There's `ExperienceData.mockWithForm()` below, and that assembles an experience model with the `"Form label"`. So change Mocks.swift to be something like `Form label - with hyphen` and then you can match that in this test.

andretortolano: See if that makes sense

iujames (approved):
LGTM, before the change, that test would have resulted in double `--`, as I understand.

mmaatttt (approved):

PR #589, opened 2025-03-11
Add workflow details to experience analytics

mmaatttt:


Sources/AppcuesKit/Presentation/ExperienceRendering/Dictionary+ExperienceProperties.swift
```diff
@@ -45,6 +45,14 @@ extension Dictionary where Key == String, Value == Any {
             properties["localeId"] = localeId
         }
 
+        if let workflowId = experience.context?.workflowId {
+            properties["workflowId"] = workflowId
+        }
+
+        if let workflowTaskId = experience.context?.workflowTaskId {
+            properties["workflowTaskId"] = workflowTaskId
+        }
```
iujames, line 54: im 99% sure this works correctly, as the BE will snake case to `workflow_id` -- that value is the critical one, as it causes the data to end up in the proper materialized view in the analytics DB

PR #590, opened 2025-03-12
Screen capture flow improvements

mmaatttt:
1. I've modernized the screen capture and annotation to use `UIGraphicsImageRenderer`. I've also added a check for a zero-sized capture, so now we show an error toast instead of crashing or having a confusing blank confirm screen.
2. To try and avoid getting the zero-sized capture failure, I've reworked and consolidated the scene/window logic to avoid a window with a zero-frame. I'm still not sure how it's possible, but if it is, we'll ignore a window like that in preference of another one.
    We had a couple different ways of finding the "main" window:
    1. The `topViewController()` implementation would try and find the `keyWindow`, with a fallback in case the `keyWindow` was an appcues window.
    2. Everything else (screen capture, element targeting) would look through all windows and take the first non-appcues window, which was probably too simplistic
    
   I've combined these approaches in the new `mainAppWindow` property that ignores all appcues windows and then takes the `keyWindow` or the best looking other window if the `keyWindow` is an appcues window.

This ends up being an all-around good refactor and simplification while also being more robust.

### DIFF
--- a/Sources/AppcuesKit/Presentation/Debugger/ScreenCapturing/Capture.swift
+++ b/Sources/AppcuesKit/Presentation/Debugger/ScreenCapturing/Capture.swift
@@ -95,22 +95,17 @@ extension Capture: Encodable {
 
 extension UIImage {
     func annotate(with targetRects: [CGRect]) -> UIImage {
-        let imageSize = self.size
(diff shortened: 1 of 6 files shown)


iujames (approved):
nice clean up!

PR #591, opened 2025-04-29
Add an option to control embed cache behavior and retained content

mmaatttt:
Matches Android https://github.com/appcues/appcues-android-sdk/pull/672

>By default, a frame with embed content can render over and over as the same frameId is re-registered in a situation like a recycler view. However, in some use cases this is undesired, and a subsequent render of the embed on a new view load would actually show content that was not supposed to be qualified. If a ~fragment~ _view controller_ is launching, for instance, and sometimes should qualify, but sometimes not - each new render of the view should only show current qualified experiences - nothing from cache. This can now be accomplished by setting retainContent = false on that particular frame view.

## Usage

This feature would require opting-in to set the new property, something like this, when registering the frame.
```swift
// Programmatic UIKit view creation
let frame1 = AppcuesFrameView(retainContent: false)

// Storyboard
@IBOutlet private var frame2: AppcuesFrameView!
...
frame2.retainContent = false

// SwiftUI
AppcuesFrame(appcues: appcuesInstance, frameID: "frame3", retainContent: false)
```

### DIFF
--- a/Sources/AppcuesKit/AppcuesKit.docc/FrameConfiguring.md
+++ b/Sources/AppcuesKit/AppcuesKit.docc/FrameConfiguring.md
@@ -41,5 +41,6 @@ VStack {
 * The `frameID` registered with Appcues for each frame should ideally be globally unique in the application, but at least must be unique on the screen where experience content may be targeted. 
 * Some ``AppcuesFrameView`` instances may not be visible on the screen when it first loads, if they are lower down on a scrolling page, for instance. However, when they scroll into view, any qualified content on that screen will then render into that position.
 * Using ``AppcuesFrameView`` in views with cell re-use (ex. `UITableView` or `UICollectionView`) is supported, but may require re-registering with a new `frameID` when cells are re-used, depending on your use case.
+* If you are re-registering a frame in a non cell re-use situation where you don't need the content to re-render, setting the ``AppcuesFrameView/retainContent`` parameter to `false` allows you to disable having the same content re-render.
 * When configuring settings for triggering embedded experience content, make sure that the experience is triggered on the same screen where the target `frameID` exists.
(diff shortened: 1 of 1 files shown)


Sources/AppcuesKit/AppcuesKit.docc/FrameConfiguring.md
```diff
@@ -41,5 +41,6 @@ VStack {
 * The `frameID` registered with Appcues for each frame should ideally be globally unique in the application, but at least must be unique on the screen where experience content may be targeted. 
 * Some ``AppcuesFrameView`` instances may not be visible on the screen when it first loads, if they are lower down on a scrolling page, for instance. However, when they scroll into view, any qualified content on that screen will then render into that position.
 * Using ``AppcuesFrameView`` in views with cell re-use (ex. `UITableView` or `UICollectionView`) is supported, but may require re-registering with a new `frameID` when cells are re-used, depending on your use case.
+* If you are re-registering a frame in a non cell re-use situation where you don't need the content to re-render, setting the ``AppcuesFrameView/retainContent`` parameter to `false` allows you to disable having the same content re-render.
```
iujames, line 44: good call in doc, will add to android

Sources/AppcuesKit/Presentation/UI/AppcuesFrameView.swift
```diff
@@ -37,6 +37,13 @@ public class AppcuesFrameView: UIView, StateMachineOwning {
     // Managed by the StateMachineDirectory
     internal var renderContext: RenderContext?
 
+    /// When retainContent is `true` (default), the embed frame content is cached and re-rendered through any
+    /// re-register of this frame until the next `screen_view` qualification occurs. This default behavior enables
+    /// common cell-reuse type of use cases, such as embeds in recycler views.
```
iujames, line 42: can update `recycler views` reference here to UITableView/UICollectionView

iujames (approved):

PR #597, opened 2025-07-16
Capture target elements from web views

mmaatttt:
This PR adds the ability to screen capture/element target into web views. To do this we call [`evaluateJavaScript`](https://developer.apple.com/documentation/webkit/wkwebview/evaluatejavascript(_:in:contentworld:)) to run JS query and transform the results into targetable rectangles.

`evaluateJavaScript` is asynchronous, and so I've chosen to adopt async/await instead of adding a mess of completion handlers. The blast radius of the changes is:

1. `UIView.asViewElement` becomes async, causing
    1. `AppcuesElementTargeting.captureLayout()` to become async (💥  this is a public api change), causing
         1. `ScreenCapturer.captureScreen()` to also become async. We init a `Task` here and that's it. [NOTE 1}
    2. `AppcuesElementTargeting.findMatches()` to become async, causing
        1. `AppcuesTargetElementTrait.decorate()` to become async [NOTE 2], causing
            1.  `protocol AppcuesBackdropDecoratingTrait` to make `decorate` async,
            causing all the other traits to also be async, and causing the
                1. `stepDecoratingTraitUpdater` in `TraitComposer` to become async, causing the 2 usages in
                    1.  `ExperienceStateMachine` to each be called from a `Task`

>NOTE 1: You could argue `captureScreen()` should be async as well and the `Task` implemented in it's called, but that would just increate the number of test changes, so I opted to leave it.
> NOTE 2: I wanted to do a `Task` here to avoid the rest of the changes, but crucially we need to be able to throw an error if the target element isn't found (to kick of retries, error analytics, etc).

The commits in this PR basically do those changes in reverse order to allow the commits to be as atomic as possible.

## The JS query
This query needs to be the same across iOS and Android (and RN for that matter). Currently I'm only taking elements with an `id` or an `appcues-id` property. This makes things simple, saving us from having to try and construct unique selectors. The developer experience will be similar to our existing patterns that require a bit of annotation. And from my testing with Ionic, several of the UI components out of the box have generated `id` values, so you get some stuff for free.

When translating to our selectors, I'm using the `#id` css selector as the `tag` value and putting the `appcues-id` value in the `appcuesID` field. I was originally combining them in a single CSS selector, but then in the case where an element had both and one was removed, it would break. Splitting them up keeps it more resilient. And we still have the option of augmenting our CSS selector syntax as the `tag` if needed.

## Misc

1. I noticed the safe area offsets being ignored in the Ionic app, and so I chased that down to `contentInsetAdjustmentBehavior` which we now use to conditionally adjust by the safe area.
2. This is this codebase's first foray into async/await, and so it necessitates some test changes. Also everything here is `@MainActor` annotated (including tests) because it's all UI operations.

### DIFF
--- a/Tests/AppcuesKitTests/Traits/AppcuesTargetElementTraitTests.swift
+++ b/Tests/AppcuesKitTests/Traits/AppcuesTargetElementTraitTests.swift
@@ -7,6 +7,7 @@
 //
 
 import XCTest
+import WebKit
 @testable import AppcuesKit
 
 @available(iOS 13.0, *)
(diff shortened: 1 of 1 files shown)


Sources/AppcuesKit/Presentation/Debugger/ScreenCapturing/ScreenCapturer.swift
```diff
@@ -34,9 +34,16 @@ internal class ScreenCapturer {
             return
         }
 
+        Task {
+            await doCapture(window: window, authorization: authorization, captureUI: captureUI)
+        }
+    }
+
+    @MainActor
+    private func doCapture(window: UIWindow?, authorization: Authorization, captureUI: ScreenCaptureUI) async {
```
mmaatttt, line 43: I split this to a helper function only to avoid a noisier diff caused by indenting everything a level.

iujames (approved):
looks great! appreciate the care to make the diff so easy to follow.

high level question - is there _any_ reason at all you think we might want to have a config option to disable this behavior? just a safeguard in case some app has an issue with the way the JS eval is trying to run? It seems like it should be safe as an exception is just gracefully resulting in no elements found? But wanted to double check that.

PR #604, opened 2025-09-08
Update clause decoding behavior

mmaatttt:
We've redefined the expected behavior of for incomplete/malformed clauses. Previously we'd fallback to `unknown` for unknown clause types, but now we also do if a known clause is malformed.

The reason for this is that it's possible to have an incomplete state in the mobile builder and we want to handle that as graciously as possible (imagine 4 valid checks and one incomplete one at the end: we can now properly handle the first 4 cases instead of failing entirely to decode the config).

We do still want to throw errors for very wrong data (ie we don't want any random JSON to decode to `unknown` since that's not helpful at all), so the key count still throws as does decoding clause groups since we'll have stricter control over how those are structured.

### DIFF
--- a/Sources/AppcuesKit/Data/Models/Clause.swift
+++ b/Sources/AppcuesKit/Data/Models/Clause.swift
@@ -137,10 +137,6 @@ extension Clause {
 // MARK: - Decodable
 
 extension Clause: Decodable {
-    private enum ClauseError: Error {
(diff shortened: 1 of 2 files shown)


iujames (approved):
makes sense

just one typo in the PR description that tripped me up very briefly

> but ~not~ now we also do if a known clause is malformed

